PR #17, opened 2026-04-10
use j.u.l.Handler to capture exceptions in ExceptionCaptureService

batya239:


jonnyzzz:
Thank you very much for the fix!

PR #26, opened 2026-04-25
Improving speed for iterative/sequential tests running (~x10 performance)

NordCoderd:
## Problem

Every `./gradlew :ij-plugin:test` invocation — including no-change re-runs — triggered full
Kotlin recompilation of both production and test code. A no-change TDD cycle took **28–30
seconds** instead of the 3–5 seconds Gradle's incremental build is designed to deliver.
`generateMetadata`, `compileKotlin`, and `compileTestKotlin` showed as executed on every run,
making TDD impractical.

## Root causes

**1. `LocalDateTime.now()` in the local dev version string** (`build.gradle.kts`)

The version for local builds was `$baseVersion.$localBuildCounter-$snapshotTimestamp-$gitHash` where `snapshotTimestamp` was evaluated at Gradle configuration time. Every second produced a new version → `generateMetadata` out-of-date → full `compileKotlin` rerun → ~20–25 s overhead per invocation regardless of any source change.

**2. `Random.nextInt()` in `GenerateMetadataTask`** (`buildSrc/.../GenerateMetadataTask.kt`)

The task used a random integer as the obfuscation multiplier for encoding the version string. Even if the version string were stable, a fresh random seed on every run produced a different `PluginMetadata.kt`, causing `compileKotlin` to re-run independently of cause 1.

The random seed was originally introduced to obfuscate a **time bomb expiration date** alongside the version - per-run randomness made the encoded date harder to reverse-engineer. The time bomb was removed later (commit `ab3db2ee`) but `Random.nextInt()` remained, orphaned from its original purpose.

**3. `npmBuild` NpmTask had no declared outputs** (`npx/build.gradle.kts`)

The task was missing `inputs`/`outputs` declarations after being migrated from a plain `Exec` task to `NpmTask`. Without declared outputs, Gradle has no baseline to compare and re-runs the task on every invocation.

**4. Tessdata download tasks made a network round-trip on every run** (`ocr-tesseract/build.gradle.kts`)

The `de.undercouch.download` plugin internally registers `outputs.upToDateWhen { false }`, so its tasks always execute regardless of any `outputs.upToDateWhen` added externally.
`onlyIfModified(true)` avoids re-downloading the body but still sends an `If-Modified-Since` HTTP request on every build — a GitHub round-trip that adds latency even when the 6 MB + 10 MB model files are already on disk.

### `build.gradle.kts` — timestamp only on deploy invocations

Removed the always-on timestamp. Instead, the timestamp is injected only when `deployPlugin` appears in `gradle.startParameter.taskNames`. This resolves the conflict between two legitimate requirements that caused the timestamp to be removed and re-introduced on the same day (Apr 13 2026):

- **TDD speed**: stable version → `generateMetadata` UP-TO-DATE → `compileKotlin` skipped
- **Hot-reload**: the IDE's Plugin Hot Reload compares the version in `plugin.xml` to decide whether to swap the JAR; same version on two consecutive deploys → no reload

```
./gradlew :ij-plugin:test  →  0.93.0.19999-SNAPSHOT-21522330          (stable)
./gradlew deployPlugin     →  0.93.0.19999-SNAPSHOT-20260425-154143-21522330  (unique)
```

CI builds are unaffected — they supply the version via `-Pmcp.build.version` or `BUILD_NUMBER`, bypassing this code path entirely.

### `buildSrc/.../GenerateMetadataTask.kt` — deterministic encoding

Replaced `Random.nextInt()` with `versionString.get().hashCode()` as the seed for the obfuscation multiplier. The encoding is now a pure function of the version string: identical inputs produce identical `PluginMetadata.kt`, so Gradle's file-hash UP-TO-DATE check passes on every no-change run. The obfuscation property is preserved - the multiplier still varies per version.

### `npx/build.gradle.kts` — declared inputs and outputs for `npmBuild`

Added tracking for the TypeScript sources, esbuild config, `package.json`, and the real output directory (`projectDir/dist`, where esbuild writes `dist/index.js`). Note: the output is at `npx/dist/`, not `npx/build/dist/` - esbuild's `outfile: "dist/index.js"` is relative to the working directory.

### `ocr-tesseract/build.gradle.kts` - skip download when files exist

Replaced `outputs.upToDateWhen` (overridden by the plugin) with `onlyIf`, which Gradle evaluates before the task body and skips the task — and the network request — entirely when the `.traineddata` file is already present:

```kotlin
onlyIf { !tessdataDownloadDir.get().asFile.resolve(fileName).exists() }
```

Caveat: if `tessdataVersion` is bumped, a `./gradlew clean` is required to evict the old files. CI agents always start clean, so this does not affect CI.

## Result

| Scenario                                       | Before    | After                             |
|------------------------------------------------|-----------|-----------------------------------|
| No-change re-run (`./gradlew :ij-plugin:test`) | ~28–30 s  | ~4 s                              |
| `./gradlew deployPlugin`                       | unchanged | unchanged (timestamp re-injected) |

On a no-change run, `generateMetadata`, `compileKotlin`, `compileTestKotlin`, and all downstream compilation tasks show `UP-TO-DATE`.  The only tasks that still always execute  are `initializeIntellijPlatformPlugin`, `prepareTest`, and `verifyBundledLibraries` - these are `@UntrackedTask` in the IntelliJ Platform Gradle Plugin and cannot be skipped without patching the plugin.

### DIFF
--- a/buildSrc/src/main/kotlin/com/jonnyzzz/mcpSteroid/gradle/GenerateMetadataTask.kt
+++ b/buildSrc/src/main/kotlin/com/jonnyzzz/mcpSteroid/gradle/GenerateMetadataTask.kt
@@ -13,7 +13,6 @@ import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.OutputFile
 import org.gradle.api.tasks.TaskAction
 import kotlin.math.absoluteValue
-import kotlin.random.Random
 
 abstract class GenerateMetadataTask : DefaultTask() {
     @get:Input
@@ -24,7 +23,11 @@ abstract class GenerateMetadataTask : DefaultTask() {
 
     @TaskAction
     fun generate() {
-        val baseMagic = (Random.nextInt().absoluteValue + 237) % 19456
+        // Seed from the version string so encoding is deterministic for identical inputs.
+        // A random seed would make the output differ on every run even if the version didn't
+        // change, defeating Gradle's up-to-date checks and forcing full recompilation every
+        // time (observed: +20 s per TDD iteration with a random seed).
+        val baseMagic = (versionString.get().hashCode().absoluteValue + 237) % 19456
 
         fun String.toEncodedSequence(): List<Int> {
             return reversed().map { it.code * baseMagic }
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -88,25 +88,14 @@ if (providedBuildVersion != null) {
 // of the CI counter so the shape stays <VERSION>.<counter>-...-<hash>, sorts after any
 // realistic CI run, and is obviously not an official build.
 //
-// The timestamp is injected ONLY when deployPlugin is explicitly requested. This resolves
-// the tension between two conflicting requirements:
-//   * TDD speed: stable version → generateMetadata UP-TO-DATE → compileKotlin skipped
-//   * Hot-reload: the IDE's Plugin Hot Reload checks the version string in plugin.xml;
-//     same version on two consecutive deploys → no reload. A per-second timestamp ensures
-//     every deploy is recognized as newer.
-// Checking startParameter.taskNames is safe at configuration time: the deploy tasks live
-// in :ij-plugin, which reads the version from rootProject.version after this block runs.
-val isDeployBuild = gradle.startParameter.taskNames.any {
-    it.contains("deployPlugin", ignoreCase = true)
-}
+// The version is stable across runs so generateMetadata stays UP-TO-DATE
+// compileKotlin is only re-run when sources actually change.
+// The git hash is the only freshness signal.
 val localBuildCounter = "19999-SNAPSHOT"
-val snapshotTimestamp = if (isDeployBuild) {
-    "-" + java.time.LocalDateTime.now().format(java.time.format.DateTimeFormatter.ofPattern("yyyyMMdd-HHmmss"))
-} else ""
 version = when {
     isReleaseBuild -> "$baseVersion-$gitHash"
     providedBuildVersion != null -> providedBuildVersion
-    else -> "$baseVersion.$localBuildCounter$snapshotTimestamp-$gitHash"
+    else -> "$baseVersion.$localBuildCounter-$gitHash"
 }
 val releaseNotesVersion = providers.gradleProperty("mcp.release.notes.version").orElse(baseVersion).get()
 val releaseNotesFile = layout.projectDirectory.file("release/notes/$releaseNotesVersion.md")
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,8 +1,5 @@
 @file:Suppress("HasPlatformType")
 
-import java.time.LocalDateTime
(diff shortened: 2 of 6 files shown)


jonnyzzz (changes requested):
I found two issues I would fix before merging:

- `npx/build.gradle.kts`: `npmBuild` declares `package.json`, sources, and `esbuild.config.mjs`, but not `package-lock.json` or `tsconfig.json`. A lockfile or TS config change can leave `npmBuild` UP-TO-DATE while `dist/index.js` is stale. Please add the missing inputs, at least `package-lock.json` and likely `tsconfig.json`.
- `ocr-tesseract/build.gradle.kts`: the new `onlyIf` skips by filename existence only, ignoring `tessdataVersion` / URL. Bumping `tessdataVersion` will silently reuse old `.traineddata` files until someone runs `clean`. Prefer a version-scoped download directory or another version-aware invalidation.

`gh pr checks` reports no checks for this branch, so please also run the targeted Gradle tasks for the touched modules.

NordCoderd:
@jonnyzzz Hi, I fixed:
- Problem with tesseract to have version based folder + applied atomic download to not relying on corrupted file in case of problems ( a bit more about tempAndMove - https://github.com/michel-kraemer/gradle-download-task#advanced )
- I fixed npx build with your following comments, I think better to have a fix than is not even if it will be rewritten. I think it's good approach.
- I dropped snapshot timestamp in build version and code becomes more simple and I think it's good enough approach as we discussed.

All the Gradle tasks related to touched modules / dependent modules were performed without problems, so, nothing was broken due to current changes.

jonnyzzz (commented):
Could you please check the kotlin download tasks too. There might be the same patch needed as well

buildSrc/src/main/kotlin/com/jonnyzzz/mcpSteroid/gradle/GenerateMetadataTask.kt
```diff
@@ -24,7 +23,11 @@ abstract class GenerateMetadataTask : DefaultTask() {
 
     @TaskAction
     fun generate() {
-        val baseMagic = (Random.nextInt().absoluteValue + 237) % 19456
+        // Seed from the version string so encoding is deterministic for identical inputs.
+        // A random seed would make the output differ on every run even if the version didn't
+        // change, defeating Gradle's up-to-date checks and forcing full recompilation every
+        // time (observed: +20 s per TDD iteration with a random seed).
+        val baseMagic = (versionString.get().hashCode().absoluteValue + 237) % 19456
```
jonnyzzz, line 30: I believe we have the same problem in the prompts generator. For each run a random string is created to encode each prompt. Thus a change to any prompt can trigger a massive rebuild. 

That is potentially another PR not that one.

NordCoderd:
@jonnyzzz you're right, there was the same problem as for Tessaract, but it didn't affect plugin builds and was hidden from fixing builds, so, for now, it's fixed.

All the tests in the touched module are passed.

jonnyzzz (approved):

jonnyzzz:
Thank you!

PR #117, opened 2026-06-18
devrig: the binary owns ~/.mcp-steroid/bin/devrig (self-heal each start + PATH + registration)

jonnyzzz:
## What

The **devrig binary** now fully owns its user-facing launcher `~/.mcp-steroid/bin/devrig` (POSIX) / `bin/devrig.cmd` (Windows). The install script owns none of it.

1. **Self-heal on every start** — `ensureBinLauncher` runs at the top of `Main.mainImpl2`, atomically (re)writing the launcher (temp + atomic move, rewrite-only-on-change, restores a lost `+x` bit). The wrapper records **absolute** paths and pins the bundled JDK via `DEVRIG_JAVA_HOME` (the install-dist start script honors it), so no caller deals with `JAVA_HOME`. Safe to change while in use.
2. **PATH ownership** — POSIX symlinks `bin/devrig` into a writable `$HOME` PATH dir (pure Java; never self-links, never clobbers a foreign devrig, handles relative pre-existing links). Windows registers the bin dir on the **user** PATH via one PowerShell call, **deduped to exactly one entry**, marker-gated to run once per install — and **never on the `devrig mcp` hot path** (it would delay the first serve; agents launch the wrapper by absolute path anyway).
3. **`devrig install`** registers the **stable wrapper** (not the content-addressed install tree), so upgrades need no re-registration. It force-writes the wrapper (explicit user intent) so it never registers a path that doesn't exist.
4. **Undocumented gate** `DEVRIG_BIN_NO_AUTO_REGISTER` (`yes`/`no`/`true`/`false`): default **ON for release**, **OFF for SNAPSHOT/dev/test** builds (so a dev build never clobbers the real launcher); the env value overrides either way.

## Validation

- `:npx-kt:test` (full) + Docker `:npx-kt:integrationTest` (`CliBinLauncherIntegrationTest`, `CliInstallIntegrationTest`, real dist in Linux containers) green.
- IDE inspections: **0 WARNING+** on all changed files.
- Reviewed by an internal 3-reviewer quorum (**GO**) and an independent 3× `run-agent.sh` peer review (2× claude, 1× codex) — all findings addressed (gated-install dangling wrapper, lost `+x` self-heal, relative-symlink resolution, PowerShell off the `mcp` hot path).
- **Live on Windows 10** (eugene-x220, Corretto 25): `devrig.cmd` written + idempotent; PATH dedup 3→1; concurrent rewrite of `devrig.cmd` while a `devrig mcp` launched through it is still running; `mcp` spawns no PowerShell; `install` force-writes on a SNAPSHOT dist.

### DIFF
--- a/npx-kt/src/integrationTest/kotlin/com/jonnyzzz/mcpSteroid/devrig/cli/CliBinLauncherIntegrationTest.kt
+++ b/npx-kt/src/integrationTest/kotlin/com/jonnyzzz/mcpSteroid/devrig/cli/CliBinLauncherIntegrationTest.kt
@@ -1,0 +1,82 @@
+/* Copyright 2025-2026 Eugene Petrenko (mcp@jonnyzzz.com); Copyright 2025-2026 JetBrains. Use of this source code is governed by the Apache 2.0 license. */
+package com.jonnyzzz.mcpSteroid.devrig.cli
+
+import com.jonnyzzz.mcpSteroid.testHelper.CloseableStackHost
+import com.jonnyzzz.mcpSteroid.testHelper.docker.startProcessInContainer
+import com.jonnyzzz.mcpSteroid.testHelper.process.ProcessResult
+import com.jonnyzzz.mcpSteroid.testHelper.process.assertExitCode
+import org.junit.jupiter.api.AfterEach
+import org.junit.jupiter.api.Test
+import kotlin.test.assertTrue
+
+/**
+ * End-to-end Docker coverage for the devrig binary OWNING `~/.mcp-steroid/bin/devrig` — the launcher is
+ * (re)written on EVERY start, symlinked onto PATH, and the whole chain (wrapper → DEVRIG_JAVA_HOME →
+ * install-tree launcher) actually runs. Also pins the undocumented `DEVRIG_BIN_NO_AUTO_REGISTER` gate:
+ * OFF by default on this SNAPSHOT dist, ON when the env opt-in is given.
+ *
+ * Runs the REAL devrig dist inside a throwaway Linux container (see [startDevrigCliContainer]); never on
+ * the host, which would create the developer's real `~/.mcp-steroid`.
+ */
+@Suppress("FunctionName")
+class CliBinLauncherIntegrationTest {
+    private val lifetime = CloseableStackHost()
+
+    @AfterEach
+    fun tearDown() {
+        lifetime.closeAllStacks()
+    }
+
+    @Test
+    fun `with the opt-in, every start writes bin devrig, symlinks it onto PATH, and the wrapper runs`() {
+        val devrig = lifetime.startDevrigCliContainer()
+        // One shell: make a PATH dir under HOME, opt in, start devrig once (self-heal fires), then prove
+        // the launcher was written, symlinked, and the symlinked wrapper actually launches devrig.
+        val script = """
+            set -eu
+            mkdir -p "${'$'}HOME/.local/bin"
+            export PATH="${'$'}HOME/.local/bin:${'$'}PATH"
+            export DEVRIG_BIN_NO_AUTO_REGISTER=false
+            "${devrig.launcher}" version >/dev/null 2>&1 || true
+            echo "WRAPPER_EXISTS=${'$'}([ -x "${'$'}HOME/.mcp-steroid/bin/devrig" ] && echo yes || echo no)"
+            echo "SYMLINK_TARGET=${'$'}(readlink "${'$'}HOME/.local/bin/devrig" 2>/dev/null || echo none)"
+            echo "===WRAPPER==="
+            cat "${'$'}HOME/.mcp-steroid/bin/devrig"
+            echo "===RUN_VIA_SYMLINK==="
+            devrig version 2>/dev/null | head -1
+        """.trimIndent()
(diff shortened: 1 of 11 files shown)


npx-kt/src/integrationTest/kotlin/com/jonnyzzz/mcpSteroid/devrig/cli/CliBinLauncherIntegrationTest.kt
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * End-to-end Docker coverage for the devrig binary OWNING `~/.mcp-steroid/bin/devrig` — the launcher is
+ * (re)written on EVERY start, symlinked onto PATH, and the whole chain (wrapper → DEVRIG_JAVA_HOME →
+ * install-tree launcher) actually runs. Also pins the undocumented `DEVRIG_BIN_NO_AUTO_REGISTER` gate:
+ * OFF by default on this SNAPSHOT dist, ON when the env opt-in is given.
+ *
+ * Runs the REAL devrig dist inside a throwaway Linux container (see [startDevrigCliContainer]); never on
+ * the host, which would create the developer's real `~/.mcp-steroid`.
+ */
+@Suppress("FunctionName")
```
jonnyzzz: not great, remove it, fix the warning

npx-kt/src/integrationTest/kotlin/com/jonnyzzz/mcpSteroid/devrig/cli/CliBinLauncherIntegrationTest.kt
```diff
@@ -0,0 +1,82 @@
...
+    private val lifetime = CloseableStackHost()
+
+    @AfterEach
+    fun tearDown() {
+        lifetime.closeAllStacks()
+    }
+
+    @Test
+    fun `with the opt-in, every start writes bin devrig, symlinks it onto PATH, and the wrapper runs`() {
+        val devrig = lifetime.startDevrigCliContainer()
```
jonnyzzz, line 31: container must not have JAVA installed, check it

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/InstallCommand.kt
```diff
@@ -9,23 +9,41 @@ import com.jonnyzzz.mcpSteroid.aiAgents.mcpAddStdioInvocation
...
-): Int = runInstallCommand(
-    command = command,
-    launcher = resolveDevrigLauncher(),
-    javaHome = Path.of(System.getProperty("java.home")),
-    out = mcpStdout,
-    err = System.err,
-    runner = runner,
-)
+): Int {
+    ensureBinLauncher(homePaths, force = true, registerWindowsPath = true)
```
jonnyzzz, line 31: is that correct that we are not looking for writable PATH on Windows?

npx-kt/src/integrationTest/kotlin/com/jonnyzzz/mcpSteroid/devrig/cli/CliBinLauncherIntegrationTest.kt
jonnyzzz: Removed the `@Suppress("FunctionName")`. Backtick `@Test` names are not flagged in this project (live IDE CodeSmellDetector reports 0 warnings without it) — it was a redundant copy-paste, nothing else to fix. Fixed in c7b8de15.

npx-kt/src/integrationTest/kotlin/com/jonnyzzz/mcpSteroid/devrig/cli/CliBinLauncherIntegrationTest.kt
jonnyzzz: A literal no-Java container is not achievable: the dist bundles no JDK, so the FIRST run (which writes the wrapper) needs some java — chicken-and-egg. Instead the test now proves DEVRIG_JAVA_HOME is the *sole* JDK source: after seeding the wrapper it poisons `java` on PATH with an `exit 97` stub (first on PATH) and unsets JAVA_HOME, then runs the symlinked wrapper and asserts a version still prints — only possible via the absolute `$DEVRIG_JAVA_HOME/bin/java` the start script resolves. If the pin ever regresses, the poison fires and the test fails. Docker-validated. Fixed in c7b8de15.

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/InstallCommand.kt
jonnyzzz: Correct as-is (confirmed by 3 independent reviews). Windows has no `~/.local/bin`-on-PATH convention and symlinks need admin/Developer Mode, so we register `~/.mcp-steroid\bin` on the HKCU user PATH via PowerShell (deduped to one entry, marker-gated, non-admin). POSIX scans for a *writable* dir only because it drops a symlink into an existing one — there is no equivalent to search for on Windows. Agents invoke the wrapper by absolute path, so MCP works regardless of PATH.

PR #113, opened 2026-06-15
Installer: version.json-driven, generated self-contained install scripts (v8)

jonnyzzz:
**Draft / WIP** — the dedicated PR for the self-contained installer epic (separate from all other work, as requested).

## Status
- ✅ **Design landed**: `docs/installer-v8-design.md` (v8) — version.json-driven, generated `install.sh`/`install.ps1`, `~/.mcp-steroid/{bin,binaries}` content-addressed (12-char sha, no cleanup), devrig `bin/` self-registration, daily Corretto/Azul coordinate-PR, Docker tests (sh + PowerShell-on-Linux).
- ✅ **Cross-model quorum** (3× `run-agent.sh`: Claude + Codex): 2 PASS_WITH_NOTES + 1 HOLD → resolutions **R-1…R-15** adopted (§12).
- ✅ **v8.2** (§13): `version.json` stays minimal (`{version-base}`); the generated script is self-contained, so no installer block is published; committed coordinate files are the URL source.
- ⬜ **P3 implement** · ⬜ **P4 Docker integration tests** · ⬜ **P5 impl-gate quorum**.

## Decisions (see `TASKS.md` "Epic — version.json-driven self-contained installer")
D1 version.json (minimal, v8.2) · D2 generated scripts (data baked) · D3 12-char-sha add-only cache · D4 `bin/` + devrig self-registration · D5 daily coordinate PR · D6 Windows-arm64 → **Azul Zulu 25** (Corretto has no win-arm64) · D7 macOS+Linux=`sh`, Windows=`ps1` · D8 test `ps1` via PowerShell-on-Linux Docker.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #122, opened 2026-06-18
website-gen: on-disk cache + computed JDK data model (Corretto 25 + Azul Zulu 25, PGP-verified)

jonnyzzz:
## What

Expands `:website-gen` with two intertwined capabilities, building toward #113's installer-script generation:

### 1. On-disk caching (`Cache.kt`)
A tiny keyed blob cache with a static factory:
- **`Cache.onDisk(root)`** — one file per key (sha256 of the key), atomic temp+move; root lives **outside `build/`** so the multi-hundred-MB JDK archives survive `clean` and are shared across runs.
- **`Cache.inMemory()`** — `ConcurrentHashMap`-backed, for tests (no disk).

On top:
- **`getOrCompute<T>`** (Kotlin Serialization) + **`getOrComputeBytes`**
- **`downloadWithEtag`** — HEAD `ETag` cache; **fails fast if the host exposes no ETag** (no stale-binary workaround).
- **`downloadVerifyingSha256`** — content-addressed cache for ETag-less hosts (Azul's CDN); the sha256 is both the cache key and the integrity check.

The cache root is computed in Gradle and **passed to the generator via `--cache-dir`** — the generator never guesses it.

### 2. Computed JDK data model (`JdkArtifacts.kt`)
**`resolveAllJdks(cache)` → `JdkModel`** — one fully-**computed** `JdkArtifact` (`url` / `size` / `sha256` / `javaHome`) per supported platform, nothing hand-pinned:

| Vendor | Platforms | Latest-25 resolution |
|---|---|---|
| Amazon Corretto 25 | linux glibc + musl(alpine) ×64/arm64, macOS ×64/arm64, windows-x64 | `latest` alias → versioned URL |
| Azul Zulu 25 | windows/aarch64 | Azul **Metadata API** (latest GA) |

- `javaHome` is discovered by **scanning archive entries** for `bin/java[.exe]`, so macOS `Contents/Home` and the Linux/Windows top dir are handled uniformly (the real Corretto Windows dir `jdk25.0.3_9` is computed, not guessed).
- **Vendor-natural validation for BOTH vendors** via detached **OpenPGP** signatures, verified with BouncyCastle (`PgpVerifier`): Corretto's `.sig` against the Amazon Corretto release key; Azul's Metadata-API `signature-binary` against the Azul package signing key.
- `JdkModelMain` + the `generateJdkModel` Gradle task wire it all together.

## Testing

- **18 hermetic unit tests** (synthetic tar.gz/zip archives + real BouncyCastle signing; no network), zero warnings.
- **Live end-to-end run** against real vendor sources: 8 artifacts resolved, **all OpenPGP signatures verified**, cache populated outside `build/`. Re-run is a **15s cache hit vs 2m45s cold**.


website-gen/src/main/kotlin/com/jonnyzzz/mcpSteroid/websitegen/Cache.kt
```diff
@@ -0,0 +1,191 @@
...
+
+/**
+ * Download [url] and cache it content-addressed by a KNOWN-good [sha256] (e.g. a checksum published by
+ * the vendor's metadata API). This is the right tool when the host exposes no usable HEAD `ETag` but
+ * does publish a trusted hash — the hash is both the cache key (byte-identical content -> same key ->
+ * cache hit) and the integrity check. The freshly downloaded bytes are verified against [sha256] before
+ * being stored; a cache hit needs no re-verification because the key IS the verified hash.
+ */
+fun Cache.downloadVerifyingSha256(url: String, sha256: String, http: HttpFetcher = KtorHttpFetcher): ByteArray =
+    getOrComputeBytes("sha256:${sha256.lowercase()}") {
```
jonnyzzz: include prefix like url-domain to make keys more readable

website-gen/src/main/kotlin/com/jonnyzzz/mcpSteroid/websitegen/JdkArtifacts.kt
```diff
@@ -0,0 +1,215 @@
...
+
+// ── Amazon Corretto (GPG-signed; ETag-cached) ────────────────────────────────────────────────────
+
+private const val CORRETTO_KEY_URL = "https://apt.corretto.aws/corretto.key"
+private const val CORRETTO_LATEST = "https://corretto.aws/downloads/latest"
+
+private data class CorrettoSpec(val platform: JdkPlatform, val archive: ArchiveType, val alias: String)
+
+// The `latest` aliases auto-resolve to the newest Corretto 25 build (currently 25.0.3.9.1) — no pin.
+private val CORRETTO_SPECS = listOf(
```
jonnyzzz, line 116: resolve names from the Github artifact to avoid hardcoded names

PR #140, opened 2026-06-22
Claude plugin

### LINKED ISSUE #134: [claude-plugin] Tests: verify plugin structure and scripts
## Context

Automated tests to catch regressions in the plugin's file structure, manifest,
and wrapper scripts. No Docker or real IDE needed — runs in normal Gradle check.

## Tasks

- [x] Verify `./gradlew :claude-plugin:assemble` produces the expected file set (like `:npx`'s `verifyPackageFiles` task)
- [x] Validate `plugin.json` — well-formed JSON, required fields present
- [x] Validate `.mcp.json` — well-formed JSON, correct server name, `type: stdio`
- [x] Check `devrig-start` has no stdout writes before `exec`
- [x] Check `devrig-start.cmd` uses `%USERPROFILE%` not `~`
- [x] Check both scripts delegate to devrig and don't exit early
- [x] Add `:claude-plugin:check` to the appropriate CI aggregator task

## Notes

- Blocked by: #133

### LINKED ISSUE #137: [claude-plugin] Auto-install devrig via /mcp-steroid:setup slash command
## Context

The plugin (#133) requires `devrig` (the JVM MCP bridge) to be installed at
`~/.mcp-steroid/bin/devrig`. It is **not** bundled in the plugin (binary + JDK ≈ 300 MB).

Claude Code has **no install-time / postinstall hook** — nothing runs when a user clicks
`/plugin install`. So we cannot install automatically at install time. The chosen mechanism
is a **slash command** the user runs once after installing the plugin.

**We do NOT write our own downloader.** A canonical, signed installer already exists:

- `:installer-gen` generates `install.sh` / `install.ps1` from the resolved JDK model
  (templates in `installer-gen/src/main/resources/templates/`). It bakes in every per-platform
  URL + SHA-256, downloads devrig + a matching Corretto/Azul JDK 25, **verifies SHA-256**,
  unpacks under `~/.mcp-steroid/binaries/`, and hands off to `devrig install devrig` (devrig
  then owns its own `~/.mcp-steroid/bin` launcher + PATH).
- It is **non-interactive** and does **not** touch agent configs (that is a separate explicit
  `devrig install` step) — safe to invoke from a command.
- All output goes to **stderr**; it is **idempotent** (re-run after an interrupted download just
  resumes / reports "already installed").
- It is published to `https://mcp-steroid.jonnyzzz.com/install.sh` / `install.ps1` as part of the
  release process (website `release-instructions.md`, Stage 7b/7c).

Reusing it = no duplicated platform detection / checksum / JDK logic, and we keep the signed
trust chain. A hand-written plugin installer would lose all of that.

## User flow

1. `/plugin install mcp-steroid` — plugin installed; devrig not present yet.
2. The MCP server can't start; the `/plugin` Errors tab shows a clear message:
   `devrig not found — run /mcp-steroid:setup`.
3. User runs **`/mcp-steroid:setup`** → Claude runs the official installer via Bash/PowerShell →
   progress streams **in the chat** → devrig + JDK install.
4. If interrupted (Ctrl+C, timeout, network) → non-zero exit → Claude reports
   "install was interrupted — run /mcp-steroid:setup again". The installer is idempotent, so
   re-running is safe and resumes.
5. Restart Claude → MCP server starts → tools available.

## Tasks

- [x] `commands/setup.md` — the `/mcp-steroid:setup` slash command. Instructs Claude to run the
      installer (selecting `install.sh` vs `install.ps1` by OS), show progress, check the exit
      code, and on failure tell the user to run `/mcp-steroid:setup` again.
- [x] `bin/install-devrig` (POSIX) — thin wrapper over
      `curl -fsSL https://mcp-steroid.jonnyzzz.com/install.sh | sh`. Single point we can validate
      (stderr-only, references `install.sh`, fail-fast). The command calls this rather than
      hard-coding `curl` in markdown.
- [x] `bin/install-devrig.ps1` (Windows) — equivalent wrapper over
      `irm https://mcp-steroid.jonnyzzz.com/install.ps1 | iex`. (Windows MCP startup itself is #141.)
- [x] Update `bin/devrig-start` so that when devrig is missing it prints a clear stderr message
      pointing at `/mcp-steroid:setup` and exits 1 — it does **not** install itself.
- [x] `:claude-plugin` validation: `commands/setup.md` exists + references the installer;
      `bin/install-devrig*` are stderr-only, fail-fast, and reference the canonical installer URL.
- [x] Update `expectedFiles` in `verifyPluginFiles` (+`commands/setup.md`, +`bin/install-devrig`,
      +`bin/install-devrig.ps1`).

## Non-goals (explicitly do NOT do)

- ❌ Hand-written platform detection, GitHub-Releases download, checksum verification, or JDK
      download in the plugin — all owned by `:installer-gen` + the devrig binary.
- ❌ Bundling the installer/binary inside the plugin zip.
- ❌ Auto-registering devrig with agents (the plugin's own `.mcp.json` already wires Claude up).

## Dependencies

- The canonical `install.sh` / `install.ps1` must be live for the target release (already part of
      the release flow).
- Windows end-to-end depends on **#141** (cross-platform MCP server startup); the `.ps1` setup
      branch lands with it. macOS/Linux can ship first.

## Corrections to the original issue

- The original referenced `:jdk-downloader` — no such module. JDK logic lives in `:installer-gen`
      (`CorrettoJdk.kt` / `AzulJdk.kt`).
- The original "bundle the script to avoid network at startup" rationale doesn't hold: install must
      download ~300 MB regardless, and runtime download is not a marketplace-policy problem (#136).

### LINKED ISSUE #160: /devrig:status quotes bug on Windows
Here is the warning i see on running status on Windows machine while working on #141 

devrig status: ⚠️ drift detected (exit 1)

devrig is installed (v0.101-40690055) and the launcher is present at %USERPROFILE%\.mcp-steroid\bin\devrig.cmd. The IDE backend is reachable (1 of 1). However, the mcp-steroid MCP registration for Claude has drifted.

What the check found:

- Currently registered (matched by name + config):
cmd.exe /d /c C:\Users\Grigorii.Guliuk\.mcp-steroid\bin\devrig.cmd mcp
- Canonical form devrig wants (note the quoted path):
cmd.exe /d /c "C:\Users\Grigorii.Guliuk\.mcp-steroid\bin\devrig.cmd" mcp

Repair plan the check printed:
- remove mcp-steroid
- add mcp-steroid → cmd.exe /d /c "C:\Users\Grigorii.Guliuk\.mcp-steroid\bin\devrig.cmd" mcp

The only difference is that the launch command's path isn't quoted. It's a stale launch command — harmless today, but it won't survive a path with spaces and devrig flags it as drift.

### LINKED ISSUE #201: Turn devrig bootstrap into a hot-swapping MCP proxy (no restart after install)
Today users must **restart Claude** after the background devrig download finishes: Claude is still connected to the short-lived bootstrap stdio process, whose tool list is frozen at `devrig_status`.

**Idea:** keep the bootstrap alive as a thin MCP proxy. When the download completes, it spawns `devrig mcp` as a child, proxies traffic to it, and fires `notifications/tools/list_changed` — Claude re-fetches the tools on the next turn. No restart.

Verified feasible on Claude Code 2.1.197: the `list_changed` handler is registered (re-fetches tools), gated on the server advertising `capabilities.tools.listChanged=true`.

BraindeadDogge:
Middle phase, required to apply for Claude Community Plugin


claude-plugin/.claude-plugin/plugin.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "mcp-steroid",
+  "description": "Give AI the whole IntelliJ, not just the files. Connects Claude to all running JetBrains IDEs via the devrig MCP bridge.",
+  "version": "0.0.0-dev",
```
jonnyzzz: version should update with the build. 

Would it make sense to generate these files with the website-gen project? Once included there, it will be easy to run website locally and/or deploy to the production once ready.

So far we have the plugin.xml for IntelliJ as yet another source of thuth.

Better name it devrig

claude-plugin/.claude-plugin/plugin.json
```diff
@@ -0,0 +1,12 @@
...
+  "author": {
+    "name": "Eugene Petrenko",
+    "email": "mcp@jonnyzzz.com"
+  },
+  "homepage": "https://mcp-steroid.jonnyzzz.com",
+  "repository": "https://github.com/jonnyzzz/mcp-steroid",
+  "license": "Apache-2.0"
```
jonnyzzz: we also have EULA

claude-plugin/bin/devrig-start
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+# MCP Steroid -- stdio MCP bridge launcher for macOS/Linux.
+# stdout is the MCP JSON-RPC channel: nothing must be written to it before exec, therefore all diagnostic output goes to stderr
+
+DEVRIG="$HOME/.mcp-steroid/bin/devrig"
+
+if [ ! -x "$DEVRIG" ]; then
+  echo "mcp-steroid: devrig not found at $DEVRIG" >&2
```
jonnyzzz: we can call the install script from the website here

claude-plugin/.mcp.json
```diff
@@ -0,0 +1,8 @@
+{
+  "mcpServers": {
+    "mcp-steroid": {
+      "type": "stdio",
+      "command": "${CLAUDE_PLUGIN_ROOT}/bin/devrig-start"
```
jonnyzzz: this is not cross-platform

claude-plugin/bin/devrig-start
BraindeadDogge: install script will be added in the #137

claude-plugin/build.gradle.kts
```diff
@@ -0,0 +1,237 @@
...
+// Patches the version placeholder in plugin.json and copies all plugin files
+// into build/plugin/ so the Zip task has a clean, versioned staging area.
+val preparePluginFiles = tasks.register("preparePluginFiles") {
+    group = "claude-plugin"
+    description = "Stage plugin files into build/plugin/ with patched version"
+
+    val sourceDir = projectDir
+    val outputDir = layout.buildDirectory.dir("plugin")
+
+    inputs.property("pluginVersion", pluginVersion)
```
jonnyzzz, line 43: better pass `{ version }` it will let Gradle better optimize caches

claude-plugin/build.gradle.kts
```diff
@@ -0,0 +1,237 @@
+import groovy.json.JsonSlurper
+import org.gradle.api.tasks.bundling.Zip
+import java.util.SortedSet
+
+plugins {
+    base
+}
+
+val pluginVersion = version.toString()
```
jonnyzzz, line 9: move it down inside doLast { }

claude-plugin/build.gradle.kts
```diff
@@ -0,0 +1,237 @@
...
+    outputs.dir(outputDir)
+
+    doLast {
+        val out = outputDir.get().asFile
+        out.mkdirs()
+
+        // .claude-plugin/plugin.json -- inject version
+        val pluginDir = out.resolve(".claude-plugin").also { it.mkdirs() }
+        val pluginJson = sourceDir.resolve(".claude-plugin/plugin.json").readText()
+        pluginDir.resolve("plugin.json").writeText(
```
jonnyzzz: use better website-gen or similar project for that substitution

jonnyzzz:
Thank you very must for the PR. Let's invest in the integration test. Secondly, that would be greate to include documentation on how to use dev build (and additional a patch to include the specification for the website). For now we can keep the custom plugins website (or one on the disk for local development)

jonnyzzz:
It might be in progress, great work! Here are some notes/ideas I spot
- conflicts with main  => pls rebase
- 70+ commits and 90+ changed files => let's better split into several smaller changes
- binaries in the repo -- I would avoid them, let's build the binaries as part of the build process
- binaries may require signatures and Apple notarization in production, I would double check if we can shrink these binaries to scripts (cmd (or powershell) and bash
- it can be the trick to test -- create `cmd` as script and check if windows would call cmd.exe instead or not. If the answer is yet -- we can still do scripting to start with
- you probably merged master, instead, please consider rebases, to keep your changes less mixes with master history

jonnyzzz:
@BraindeadDogge Grigory, thank you for carrying this much farther than the original plugin skeleton. I reviewed the history, not only the final diff: this branch contains the packaging and local-development documentation/tests we asked for, marketplace-policy research, Windows launch and quoting work, real installer coverage, progress and recovery ideas, a hot-swap prototype, cwd routing, and first-run/help guidance. PR #140 remains the implementation source for #308; I do not want that work or its authorship lost.

The branch is now 78 commits / 97 files, and `main` has moved 158 commits ahead. Asking you to rebase and split the entire branch would put the repository-integration churn on you, so I am withdrawing that as the next action. Please pause history-cleanup/rebase work on #140.

We have now made a maintainer architecture decision for #308/#307: one deterministic build will produce a cross-agent Claude/Codex integration over the same OS-installed devrig. The plugin is a thin adapter; devrig owns installation, runtime, updates, rollback, and the stable launcher. That means we will not carry forward the committed bootstrap binaries or the plugin-owned Go proxy/installer state, but we will preserve the valuable product behavior, tests, copy, and platform knowledge from this branch.

Maintainers will own:

1. Keeping this PR open as the reference branch and adding a `port / already on main / follow-up / rejected` ledger.
2. Creating clean, small PRs from current `main` and resolving conflicts:
   - foundation: dual Claude/Codex manifests, shared MCP/skills/assets, version generation, docs, validation, and the #253 script-only launcher;
   - first success: explicit canonical setup, status/doctor/help, one runnable verified task, concise recovery and compaction guidance;
   - follow-ups: only evidence-backed devrig-owned lifecycle work, plus independent cwd/installer product PRs.
3. Auditing the installer changes against current `main` and porting only genuine gaps. No committed platform binaries; no silent background install; no global `~/.mcp-steroid` uninstall.
4. Owning Claude/Codex Windows/Linux/macOS testing, CI, marketplace validation, and submission.
5. Preserving your authorship by cherry-picking your commits where practical and explicitly crediting rewritten extractions and release notes.

The detailed recovery plan and file-by-file ledger are on #308: https://github.com/jonnyzzz/mcp-steroid/issues/308

Could you reply with only:

- the **three user-visible behaviors** from this branch that you consider essential not to lose; and
- whether you would prefer to **own one small extracted PR**, **review the extractions only**, or **step back for now**?

Any of those is completely fine. There is no request to rebase 97 files. If you want a bounded slice, the foundation package review or Windows smoke verification would be especially valuable. We will ask for your review after each small extraction rather than after another large rewrite, and we will keep #140 open until the ledger and extracted PR links account for the material work.

Thank you again—this branch identified and solved much of the first-run problem that the shared Claude/Codex integration now depends on.

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/InstallCommand.kt
```diff
@@ -12,8 +12,8 @@ import java.io.PrintStream
 import java.nio.file.Path
 import kotlin.io.path.isRegularFile
 
-private const val DEVRIG_MCP_SERVER_NAME = "mcp-steroid"
-private const val DEVRIG_LEGACY_SERVER_NAME = "devrig"
+private const val DEVRIG_MCP_SERVER_NAME = "devrig"
```
jonnyzzz, line 15: **blocker — backward compatibility: the MCP server name flips, so every tool identifier changes.**

`DEVRIG_MCP_SERVER_NAME` / `DEVRIG_LEGACY_SERVER_NAME` are swapped here and `DEFAULT_SERVER_NAME` becomes `"devrig"` in `ai-agents/.../McpServerConfig.kt:6`. Existing installs get the `mcp-steroid` entry removed and re-added as `devrig`, so tools rename `mcp__mcp-steroid__steroid_*` -> `mcp__devrig__steroid_*`.

Breaks: `permissions.allow` entries in `~/.claude/settings.json`, `--allowedTools mcp__mcp-steroid__*` in CI scripts, and any CLAUDE.md/prompt that names the tool. In headless `-p` runs the call is blocked, not re-prompted.

The rename is also not propagated inside this PR:
- `npx-kt/.../McpServerListReview.kt:17` (new file in this PR) still says "`mcp-steroid` is canonical; `devrig` is a legacy/alternative spelling".
- `prompts/src/main/prompts/mcp-steroid-info.md:32` and `prompts/src/main/prompts/ide/find-duplicates.md:39` still tell agents to `ToolSearch(query="select:mcp__mcp-steroid__steroid_...")`. That tip exists to avoid `InputValidationError` — after this rename it guarantees one.

Fix: keep `mcp-steroid` for existing installs and use the new name only for fresh plugin installs, or ship an explicit migration notice in `install --check` output + release notes, and update the two prompt files and `McpServerListReview` KDoc in the same change.

jonnyzzz (commented):
## Quorum review — 5 reviewers (correctness, security, design, tests, product/API-fit), same evidence, findings deduped and verified against the source at `b49a3a3`

Scope note: this is a review of the code as written. Per the maintainer plan on #308 this branch is not merging as-is — so the point of the list below is which defects **must not be inherited by the extracted PRs**.

CI: **no checks reported** on `claude-plugin`. See blocker 3 — that is not just a missing pipeline, the new validation tasks are unreachable by design.

### Blockers
1. `npx-kt/.../InstallCommand.kt:15` — MCP server name flips `mcp-steroid` -> `devrig`, renaming every tool (`mcp__mcp-steroid__*` -> `mcp__devrig__*`). Breaks existing `permissions.allow` entries and `--allowedTools` in CI. Not propagated: `McpServerListReview.kt:17` (new in this PR) still calls `devrig` "legacy", and `prompts/.../mcp-steroid-info.md:32` + `prompts/.../ide/find-duplicates.md:39` still tell agents to load the old names.
2. `ij-plugin/.../ServerUrlWriter.kt:51` — `parentScope.coroutineContext + SupervisorJob()` makes a **root** job, and `dispose()` (:199) never cancels it. The heartbeat outlives disposal and re-creates the marker `dispose()` just deleted; the new test leaks a loop that writes into the real `$HOME` after `user.home` is restored. `UpdateChecker.kt:146` does `scope.cancel()`.
3. `build.gradle.kts:314` — `ciDistributionChecks` (the only thing that runs `:claude-plugin:check`) is referenced nowhere. All ~20 `validate*` tasks are dead on arrival.

### Major
4. `claude-plugin/bin/install-devrig:30` — `status=$?` after `curl | sh` is `sh`'s code; a failed download reports success on the upgrade path and clears the failure marker. Same masking at `DevrigSetup.kt:51`/`:159`.
5. `npx-kt/.../ConnectCommand.kt:35` — writes `extraKnownMarketplaces` + `enabledPlugins` straight into `~/.claude/settings.json`, enabling a floating-branch GitHub marketplace whose plugin runs 4 command hooks — reached from an IDE button labelled "Enable".
6. `npx-kt/.../ConnectCommand.kt:60` — SessionStart hook fires `installPlugin?action=install` at every listener on 63342-63361, with an `Origin` header added (per its own comment) to get past IntelliJ's host-trust dialog.
7. `ij-plugin/.../ServerUrlWriter.kt:120` — the marker carries the bridge bearer token and the `_ijt` token, written with default 0644; the new 10 s heartbeat re-creates it, so `chmod 600` no longer sticks.
8. `mcp-steroid-server/.../CommonToolParams.kt:29` — `markRequired()` sets the schema flag but drops parse-time enforcement; 4 of 5 tools lost the guard, so a missing `project_name` now surfaces as `Project not found: ""`.
9. `claude-plugin/bin/offer-ide:22` — Windows layout fakes `code=0` without running `connect ide`, then burns the permanent `ide-offered` marker.
10. `claude-plugin/hooks/hooks.json:20` — `check-devrig` and `offer-ide` have no cmd half (unlike `devrig-recover`/`devrig-progress`), so the whole Windows first-run nudge is silent; the polyglots are also registered extension-less.
11. `ij-plugin/.../DevrigStatusBarWidget.kt:162` — fresh `DevrigSetupRunner()` per click, no in-flight guard: two concurrent ~611 MB installers, corrupted progress, failure marker races.
12. `ij-plugin/.../DevrigSetup.kt:51` — third divergent install implementation (curl-only, no `wget` fallback; weaker installed-check than the shell path).
13. `ij-plugin/.../DevrigSetup.kt:127` — `DevrigSetupRunner` has no test at all; the streaming line-splitter and the four-way failure `reason` are unverified. `ConnectIde.kt:27` shows the seam pattern that is missing here.
14. `claude-plugin/commands/status.md:23` (+ `uninstall.md:27`) — Windows `%USERPROFILE%` and `Remove-Item` handed to the Bash tool; the uninstall delete silently does nothing after the user confirms.
15. `claude-plugin/commands/uninstall.md:29` — uninstall is not the inverse of install: plugin and settings.json keys stay, leaving a permanently failing MCP server and a nag every session.
16. `claude-plugin/build.gradle.kts:639` — the build script regex-greps Kotlin source in two other modules; a legal dedup refactor breaks the build. Belongs in a test (`NoForbiddenPluginImportsTest` is the existing idiom).
17. `ij-plugin/src/test/.../InstallerProgressTest.kt:33` — the file claims to fail when the installer templates change wording; it hardcodes the same literals and reads no template, so it protects against nothing.

### Minor / nit
18. `ij-plugin/.../OnboardingDecision.kt:89` — `Path.of` on raw PATH entries throws on quoted Windows entries; the exception is swallowed and the widget hangs on "Checking..." for the session.
19. `npx-kt/.../DevrigWorkspaceRoots.kt:32` — the degraded cwd-only fallback is cached forever; `invalidate()` has no caller.
20. `claude-plugin/bin/install-devrig.ps1:32` — `exit 1` inside `Invoke-Expression` bypasses the `catch`, so the Windows failure marker is never written.
21. `claude-plugin/bin/devrig-recover:18` — signature grep runs on the raw payload with no `isError` check; repo content can trigger RECOVERY steering after a successful call.
22. `claude-plugin/hooks/hooks.json` — all four commands interpolate `${CLAUDE_PLUGIN_ROOT}` unquoted; a space in the path word-splits and every hook goes dead silently. `commands/setup.md:19` already quotes it.
23. `npx-kt/.../ClaudePluginConnect.kt:38` and `ij-plugin/.../OnboardingDecision.kt:104` — `isClaudePluginEnabled` duplicated across two modules with duplicated test suites, and both KDocs still document the old key `devrig@mcp-steroid` (actual: `devrig@jonnyzzz`).
24. `ij-plugin/.../DevrigSetup.kt:42` — `~/.mcp-steroid` layout hand-built instead of `PidMarker.markerDirectory`, which this module already depends on; `devrigInstalled` re-derives the bin path 10 lines from `devrigBinPath`.
25. `claude-plugin/build.gradle.kts:194` — `install-devrig` is the only `bin/` script never executed by a `*Runs` task; the substring greps are satisfied by its header comment alone. Also `bin/install-devrig:10` still has `# TODO: Probably temp, i'll think about...`.
26. `ij-plugin/.../DevrigStatusBarWidget.kt:69` — `getText`/`getTooltipText` are in a private inner class and untested; `installedVersion == null` renders "devrig  is behind 0.101" (the popup path handles the same null at `DevrigWidgetPopupContent.kt:54`).
27. `build.gradle.kts:308` — the `ciDistributionChecks` KDoc is copy-pasted from the prompts aggregator and describes the wrong thing.
28. `TODO.md:13` — stale in the same commit; all five listed gaps are implemented here, one is deliberately rejected, and the line citations are wrong.

Inline comments carry the failure scenario and suggested fix for each.

ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/server/ServerUrlWriter.kt
```diff
@@ -15,39 +14,75 @@ import com.jonnyzzz.mcpSteroid.McpSteroidServerInfo
...
+ * and re-created mid-session (the devrig installer re-creates the tree; a user
+ * may `rm -rf ~/.mcp-steroid`). Without the heartbeat the marker would stay gone
+ * until the next IDE restart, and a running-IDE fast path (bootstrap Tier 1)
+ * would never find this IDE. The heartbeat re-creates it within [MARKER_HEARTBEAT].
  */
 @Service(Service.Level.APP)
-class ServerUrlWriter : Disposable {
+class ServerUrlWriter(parentScope: CoroutineScope) : Disposable {
     private val log = thisLogger()
+    private val scope = CoroutineScope(parentScope.coroutineContext + SupervisorJob() + Dispatchers.IO)
```
jonnyzzz, line 51: **blocker — the heartbeat scope is not a child of `parentScope`, and `dispose()` never cancels it.**

`parentScope.coroutineContext + SupervisorJob()` *replaces* the parent Job with a new root Job. Cancelling `parentScope` does not touch this scope.

Consequences:
1. Production: on plugin unload / `Disposer.dispose`, `dispose()` (:199) deletes the marker, the `while (isActive)` loop at :72 keeps running, and ~10 s later `writeMarkerFile` re-creates a marker for an IDE whose MCP server is gone. devrig then discovers and dials a dead endpoint until the JVM exits. The comment at :199 ("The injected scope is cancelled by the platform on dispose") is false for this scope shape.
2. Tests: `ServerUrlWriterTest.testHeartbeatRecreatesDeletedMarker` calls `scope.cancel()` on its own scope, which does not stop the leaked loop; `withTemporaryUserHome`'s `finally` then restores the real `user.home`, and `writeMarkerFile` re-reads `System.getProperty("user.home")` every tick — so the leaked 30 ms loop starts writing markers into the developer's/CI agent's real `$HOME` and running `cleanupStaleMarkers` there. It also races the temp-dir cleanup, which can throw `DirectoryNotEmptyException` (flaky test).

Sibling services with the identical scope construction do cancel: `UpdateChecker.kt:146-148` and `SteroidsMcpServer.dispose()`. Fix: `scope.cancel()` in `dispose()`, and/or parent the job properly (`SupervisorJob(parentScope.coroutineContext.job)` / `parentScope.childScope(...)`).

ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/server/ServerUrlWriter.kt
```diff
@@ -77,24 +112,20 @@ class ServerUrlWriter : Disposable {
             intellijMcpServer = IntelliJMcpServerProbe.getInstanceOrNull()?.probe(),
         )
         val content = PidMarkerJson.encode(marker)
-        log.info("Writing MCP Steroid marker (pid=$pid)\n$content")
 
         try {
+            val wasMissing = !Files.exists(file)
             Files.createDirectories(markerDir)
             cleanupStaleMarkers(markerDir)
             Files.writeString(file, content)
```
jonnyzzz, line 120: **major — the marker holds live credentials and is written with default (0644) permissions; the new heartbeat makes that unfixable by the user.**

The marker content built above includes the devrig bridge bearer token (`bearerHeaders(NpxBridgeService.getInstance().token)`, :91) and IntelliJ's `_ijt` built-in-web-server token (:147-155). `Files.createDirectories(markerDir)` + `Files.writeString(file, content)` use default umask — 0644 in a 0755 home on a shared Linux host / build agent / container.

Any other local account reads `~/.mcp-steroid/markers/<pid>.mcp-steroid` and gets a bearer token that authorises `steroid_execute_code` — arbitrary Kotlin inside the IDE JVM as the IDE user.

This PR makes it worse than `main`: previously the marker was written once at startup, so `chmod 600` stuck until restart. With `markerHeartbeatMs = 10_000` (:213) the CREATE+TRUNCATE rewrite restores default permissions within 10 seconds, every time.

Fix: create dir and file with explicit POSIX permissions (`PosixFilePermissions.asFileAttribute` — `rwx------` / `rw-------`, ACL equivalent on Windows) and re-assert them on each heartbeat rewrite.

build.gradle.kts
```diff
@@ -300,6 +301,22 @@ val ciBuildPluginTests by tasks.registering {
...
+ * Aggregator for distribution-artifact subprojects that use `:check` instead of `:test`
+ * (no compiled code, just file-structure and script-content validation).
+ *
+ * Currently covers:
+ * * `claude-plugin` — validates plugin.json, the install/hook wrapper scripts, and the setup command.
+ *
+ * These modules are excluded from [ciBuildPluginTests] (they have no `:test` task) and
+ * run here instead on a lightweight CI step.
+ */
+val ciDistributionChecks by tasks.registering {
```
jonnyzzz, line 314: **blocker — `:claude-plugin:check` is invoked by nothing, so all ~20 new `validate*` tasks are dead on arrival.**

`grep -rn ciDistributionChecks` over the repo returns exactly one hit: this declaration. Nothing runs it:
- `.github/workflows/compile-check.yml:69` (the PR gate) runs only `./gradlew compileAllClasses`
- `.github/workflows/build-plugin.yml:114` runs `buildPluginOnCI`
- `ciBuildPluginTests` explicitly excludes `claude-plugin` (:208)

The PR shows **0 checks reported** on the branch, which is consistent. So `validateMarketplaceJson`, `validateCheckDevrigRuns`, `validateOfferIdeRuns` etc. would never catch a broken launcher script or a manifest key mismatch before it reaches users.

Also: the KDoc immediately above this task (:308-313) is copy-pasted from the prompts aggregator — it talks about "the prompt-resource pipeline" and `ciBuildPromptsTests`, which is not what this task does.

Fix: add `:claude-plugin:check` to an aggregator CI already invokes (it is cheap), or add the GH/TeamCity step calling `ciDistributionChecks` in this PR, and fix the doc comment.

claude-plugin/bin/install-devrig
```diff
@@ -0,0 +1,54 @@
...
+  printf '%s\n' "$1" > "$FAILED" 2>/dev/null || true
+  echo "devrig: $1" >&2
+  exit 1
+}
+
+echo "devrig: installing devrig via $INSTALL_URL" >&2
+
+if command -v curl >/dev/null 2>&1; then
+  set +e
+  curl -fsSL "$INSTALL_URL" | sh >&2
```
jonnyzzz, line 30: **major — `status=$?` after `curl | sh` captures `sh`'s exit code, not `curl`'s, so a failed download can be reported as a successful install.**

`sh` reading an empty stdin exits 0. So when `curl -f` fails (404, blocked host, proxy, offline), `status=0` and the `fail` on :33 is skipped. On a **fresh** machine the `[ ! -x "$DEVRIG" ]` guard at :46 still catches it — but on the **upgrade** path the old launcher is still there, so :51 deletes `markers/bootstrap-install.failed` and :53-54 print "devrig installed" with exit 0. Anyone able to block or 404 the install URL can pin a user to an outdated devrig indefinitely while `/devrig:setup` reports success. Same for the `wget -qO- | sh` branch at :36.

The IDE path has the identical masking: `DevrigSetup.kt:51` builds `/bin/sh -c "curl -fsSL ... | sh"` and `:159` computes `ok = !isTimeout && exitCode == 0 && installed`.

Fix: download to a temp file and check `curl` separately (`curl -fsSL "$INSTALL_URL" -o "$tmp" || fail ...; sh "$tmp"`), or `set -o pipefail` where available; and make the post-check compare the installed *version* rather than merely testing that some file exists.

claude-plugin/bin/offer-ide
```diff
@@ -0,0 +1,48 @@
...
+# Offer at most once per machine.
+if [ -f "$MARKER" ]; then exit 0; fi
+
+# Run the offer; devrig decides (already-connected / no-IDE / install dialog / browser / manual) and
+# prints its narration to stderr. We don't parse it -- we only surface a one-line nudge to the agent.
+if [ -x "$DEVRIG" ]; then
+  "$DEVRIG" connect ide >&2 2>&1
+  code=$?
+else
+  code=0
```
jonnyzzz, line 22: **major — on the Windows layout this fakes `code=0` without ever running `devrig connect ide`, and then burns the once-per-machine marker.**

:11 lets execution continue when only `devrig.cmd` exists (which is all `ensureBinLauncherCore` writes on Windows). :18's `[ -x "$DEVRIG" ]` is then false, so this `else` sets `code=0`, the `0|1` arm at :31 writes `~/.mcp-steroid/markers/ide-offered`, and the hook prints additionalContext claiming devrig "checked for a running JetBrains IDE ... and, if it found one, offered to install it ... Do not repeat this offer."

Net effect: the IDE-plugin offer is never made on any Windows machine, the marker is permanent so it is never retried, and the agent is told a check happened that did not. This directly contradicts the comment at :25-29 ("Burn the once-per-machine marker ONLY for the outcomes devrig actually reported").

Fix: pick the launcher the same way :6-11 detect it and only write the marker after a real exit code from a real invocation.

claude-plugin/hooks/hooks.json
```diff
@@ -0,0 +1,44 @@
...
+          }
+        ]
+      }
+    ],
+    "SessionStart": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "${CLAUDE_PLUGIN_ROOT}/bin/check-devrig"
```
jonnyzzz, line 20: **major — `check-devrig` and `offer-ide` are POSIX-sh-only, so the entire Windows first-run path is silent.**

`bin/devrig-progress` and `bin/devrig-recover` start with the `:;` sh/cmd polyglot header and carry full `rem ===== Windows cmd =====` halves. `bin/check-devrig:1` and `bin/offer-ide:1` are plain `#!/bin/sh` with no cmd half — and unlike `devrig-mcp`, which needed a `devrig-mcp.cmd` sibling, they have no `.cmd` form either.

Failure: a Windows user installs the plugin without devrig. The `devrig` MCP server fails to start (nothing at `~/.mcp-steroid`), and `check-devrig` — the only thing that would say "Run /devrig:setup" — cannot run. The user sees a broken MCP server and no instruction. `offer-ide` likewise never runs.

Related, worth confirming on a real Windows box: the polyglot files are registered here **without an extension**, and cmd.exe resolves executables via `PATHEXT` — an extension-less file matches nothing. If Claude launches hooks through cmd on Windows, the cmd halves of `devrig-recover`/`devrig-progress` are unreachable too. The fact that `devrig-mcp` needed a `.cmd` sibling suggests it is.

Also: `claude-plugin/.gitattributes` pins `eol=lf` only for `devrig-mcp.cmd` and `devrig-recover`. `check-devrig`, `offer-ide`, `devrig-progress` and `install-devrig` all start `#!/bin/sh` and break under a CRLF checkout.

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/ConnectCommand.kt
```diff
@@ -0,0 +1,77 @@
...
+ * (idempotent — see [enableClaudePluginInSettings]). Writes atomically (temp file + ATOMIC_MOVE) so a
+ * concurrent Claude read never sees a torn file. Best-effort narration to [out]; failures throw.
+ */
+fun runConnectClaude(settingsPath: Path, out: PrintStream, err: PrintStream): Int {
+    val current = if (settingsPath.isRegularFile()) settingsPath.readText() else null
+    if (isClaudePluginEnabled(current)) {
+        out.println("Claude Code plugin '$CLAUDE_DEVRIG_PLUGIN_KEY' is already enabled in $settingsPath.")
+        return 0
+    }
+    val updated = enableClaudePluginInSettings(current)
```
jonnyzzz, line 35: **major — this registers a remote GitHub plugin marketplace and auto-enables its hook-executing plugin by writing `~/.claude/settings.json` directly, instead of going through Claude's own plugin-install/trust path.**

`enableClaudePluginInSettings` inserts `extraKnownMarketplaces.jonnyzzz = {source: github, repo: jonnyzzz/mcp-steroid}` and `enabledPlugins["devrig@jonnyzzz"] = true` (`ClaudePluginConnect.kt:61-73`), then atomically replaces the file.

The reachable path matters: the IDE notification button labelled "Enable" (`DevrigOnboardingService.kt:101` -> `DevrigSetup.kt:87 connectClaudeArgv`) triggers it. From the next Claude session on, that repo's `main` branch supplies `claude-plugin/hooks/hooks.json` — four `"type": "command"` hooks running on every session and every prompt. Whoever can push to `jonnyzzz/mcp-steroid` gets shell execution on every installed user's next session. The user pressed a button labelled "Enable" in an IDE; they were never shown what it grants.

Two secondary effects: the marketplace is pinned to a floating branch rather than a tag/commit, and the whole `settings.json` is reparsed with `isLenient = true` and re-emitted by `prettyJson`, silently reformatting the user's file.

Fix: shell out to Claude's own marketplace-add / plugin-enable path so its confirmation runs, or at minimum print the exact keys being written and require explicit consent (`--yes`) before touching `~/.claude/settings.json`; and pin the marketplace to a tag.

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/ConnectCommand.kt
```diff
@@ -0,0 +1,77 @@
...
+    runBlocking(Dispatchers.IO) {
+        val discovered = portDiscovery.stateSnapshot()
+        // Count of live plugin-bearing IDEs (marker-based, no network probe); >0 means the bridge already works.
+        val pluginMarkerCount = ideDiscovery.stateSnapshot().size
+        val client = InstallPluginClient { url ->
+            val response = commandHttpClient.get(url) {
+                // /api/installPlugin trusts localhost callers (RestService.isHostTrusted -> isLocalhost);
+                // devrig is a localhost tool, so we declare our real localhost origin rather than spoof a
+                // marketplace one. Without any Origin the endpoint blocks on an "unknown host" dialog.
+                header(HttpHeaders.Origin, "http://127.0.0.1")
```
jonnyzzz, line 60: **major — an unsolicited `SessionStart` hook sends `installPlugin?action=install` to every listener on 63342-63361, with an `Origin` header added specifically to get past IntelliJ's host-trust dialog.**

`bin/offer-ide:19` runs `devrig connect ide` on session start with no user request. `runConnectIde` (`ConnectIde.kt:109-122`) then loops over every discovered built-in-server port issuing `checkCompatibility` and `action=install` for `com.jonnyzzz.mcp-steroid` — including IDE windows opened for unrelated work, and any non-JetBrains service holding a port in that range.

The comment on :57-59 is explicit that the `Origin` header exists because "without any Origin the endpoint blocks on an 'unknown host' dialog" — i.e. a trust check is deliberately satisfied so an unattended request goes through. If the install dialog is dismissed, `browser.open(MARKETPLACE_INSTALL_URL)` opens a browser page, also unprompted.

Fix: make `connect ide` an explicit user action (a `/devrig:connect-ide` command or a confirmed IDE action) rather than a SessionStart side effect, and only target ports whose `/api/about` probe positively identified a JetBrains IDE.

mcp-steroid-server/src/main/kotlin/com/jonnyzzz/mcpSteroid/server/CommonToolParams.kt
```diff
@@ -12,12 +13,21 @@ import com.jonnyzzz.mcpSteroid.mcp.string
...
+    fun projectName(requiredInSchema: Boolean = true): InputSchemaElement<String?> {
+        val base = InputSchemaElement.param("project_name")
+            .description(
+                if (requiredInSchema) "Project name (from steroid_list_projects)"
+                else "Project name. Omit to auto-detect from your working directory; pass it " +
+                    "(from steroid_list_projects) only to disambiguate or override."
+            )
             .string()
-            .required()
+        return if (requiredInSchema) base.markRequired() else base
```
jonnyzzz, line 29: **major — `markRequired()` sets the schema flag but drops the parse-time enforcement `required()` provided, and 4 of the 5 call sites did not re-add a guard.**

`McpSchema.kt:168` only stamps `required = true` into the emitted schema; `required()` (:171-182) is what throws `ToolCallErrorException("Parameter project_name of type string is required")`. So on the in-IDE surface (`projectNameRequired = true`) the declared contract and the enforced contract now disagree.

Failure: a client calls `steroid_execute_code` without `project_name`. Previously: a precise missing-parameter error. Now `context[projectName].orEmpty()` yields `""`, which reaches `ProjectScopedToolHandler.resolveProject("")` and surfaces as `Project not found: "". Available projects: [...]` — the agent is told its project name is wrong rather than missing, and will retry with a different name.

Affected: `ExecuteCodeTool.kt:129`, `VisionInputTool.kt:66`, `VisionScreenshotTool.kt:44`, `FetchResourceToolHandler.kt:65`. Only `ExecuteFeedbackTool.kt:81-83` re-added `if (projectNameRequired && projectName.isBlank())`. Note `taskId()` and `reason()` in this same file still use `.required()`, so two required params of one tool now behave differently.

Fix: return the enforcing `required()` element when `requiredInSchema` is true and use `markRequired()` only on the optional devrig path — or add the blank guard to all five `call()` bodies.

ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/onboarding/DevrigStatusBarWidget.kt
```diff
@@ -0,0 +1,168 @@
...
+    private fun escape(text: String): String = StringUtil.escapeXmlEntities(text)
+
+    private fun runWidgetAction(action: DevrigWidgetAction) {
+        analyticsBeacon.capture(
+            "devrig_onboarding_action",
+            project,
+            mapOf("action" to "widget", "widget_action" to action.name),
+        )
+        when (action) {
+            DevrigWidgetAction.INSTALL, DevrigWidgetAction.UPDATE -> DevrigSetupRunner().runEnable(project)
```
jonnyzzz, line 162: **major — a fresh `DevrigSetupRunner()` per click with no in-flight guard: two concurrent ~611 MB installers can run against the same `~/.mcp-steroid` tree.**

`DevrigOnboardingService.kt:35` holds its own `DevrigSetupRunner()`, and this line builds another one on every widget click. Neither has a re-entrancy guard (`DevrigSetup.kt:66-69`).

Failure: the user clicks "Enable" on the onboarding balloon, starting install #1. The widget presentation is only recomputed in the runner's `finally`, so it still reads "devrig: not connected"; the user clicks it and presses Install -> install #2. Both write `.tmp.*` into `~/.mcp-steroid/binaries`, and `stagedBytes` (`DevrigSetup.kt:214`) globs `".tmp.*"` and therefore sums *both* runs, so each progress fraction is wrong. Whichever finishes last decides whether `markers/bootstrap-install.failed` is written or cleared — a failed run can erase a successful run's clean state. A second IDE window gives the same double-entry surface.

Fix: make `DevrigSetupRunner` an `@Service(Service.Level.APP)` (per `ij-plugin/CLAUDE.md`: "Use IntelliJ services instead of object/singletons") with an `AtomicBoolean` in-flight guard, like `DevrigOnboardingService.offered`; a second Enable/Install should just focus the running task.

ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/onboarding/DevrigSetup.kt
```diff
@@ -0,0 +1,318 @@
...
+    userHome.resolve(".mcp-steroid").resolve("markers").resolve("bootstrap-install.failed")
+
+/** Where the installer stages its in-flight downloads (`.tmp.*` under `binaries/`). */
+private fun devrigBinariesDir(userHome: Path): Path =
+    userHome.resolve(".mcp-steroid").resolve("binaries")
+
+/** Argv that runs the canonical devrig installer for this OS (the same one-liner the docs publish). */
+fun installerArgv(windows: Boolean): List<String> =
+    if (windows) listOf("powershell", "-NoProfile", "-ExecutionPolicy", "Bypass", "-Command", "irm $INSTALL_PS1_URL | iex")
+    else listOf("/bin/sh", "-c", "curl -fsSL $INSTALL_SH_URL | sh")
```
jonnyzzz, line 51: **major — third divergent implementation of the install one-liner; the IDE path has no `wget` fallback and a weaker installed-check than the shell path.**

`claude-plugin/bin/install-devrig:28-42` tries `curl`, then falls back to `wget -qO- "$INSTALL_URL" | sh`. This line is curl-only.

Failure: on a minimal Debian/container host that ships `wget` but not `curl`, `/devrig:setup` from Claude Code succeeds while clicking "Enable" in the IDE fails — and because the pipeline's exit status is `sh`'s 0, the user gets the misleading "the installer finished but devrig was not found at ~/.mcp-steroid/bin/devrig", written as the shared failure marker.

Second divergence: `OnboardingDecision.kt:42-45` uses `Files.isRegularFile`, `bin/install-devrig:46` uses `[ ! -x "$DEVRIG" ]`. A launcher that lost its exec bit makes the IDE say "installed" and the shell say "failed" about the same file.

Fix: have the IDE path invoke `bin/install-devrig` (or a single wrapper generated by `:installer-gen`, which already owns `install.sh.tmpl`) instead of re-spelling the one-liner, and align the installed-check.

ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/onboarding/DevrigSetup.kt
```diff
@@ -0,0 +1,318 @@
...
+                } finally {
+                    // Whatever happened, the widget must reflect reality afterwards.
+                    DevrigConnectionStateService.getInstance().refreshAsync()
+                }
+            }
+        })
+    }
+
+    /** Runs the canonical installer with live progress. Returns true when devrig is installed afterwards. */
+    private fun installDevrig(
```
jonnyzzz, line 127: **major — `DevrigSetupRunner`, the class that actually performs the install, has no test at all.**

`grep DevrigSetupRunner ij-plugin/src/test/` returns nothing; `DevrigSetupTest.kt` is 34 lines covering three pure helpers (`installerArgv`, `connectClaudeArgv`, `devrigBinPath`).

Unverified branches that matter:
- the four-way failure `reason` at :174-179 and the failure-marker write it drives (including "exit 0 but devrig missing", which is exactly the case that would otherwise be silent);
- `runCommandStreaming`'s newline buffering at :266-274 — the thing this PR was written to fix, and where an off-by-one swallows the last `[mcp-steroid]` line or splits a UTF-8 chunk;
- `startDownloadPoller`'s `.tmp.*` glob and divide-by-`expected` at :200-214, which silently reports 0% if the installer template renames its staging prefix.

The seams exist elsewhere in this PR — `ConnectIde.kt:27-30` defines `fun interface InstallPluginClient` "so tests inject a fake instead of a live Ktor client", giving `ConnectIdeTest.kt` full coverage of `runConnectIde`. Here `runCommand`/`runCommandStreaming` (:235, :251) hard-wire `ExecUtil` / `OSProcessHandler` with no injection point.

Fix: extract the line-splitting loop and the `reason` selection into pure top-level functions next to `parseInstallerLine` and unit-test them; or introduce a `CommandRunner` seam like `InstallPluginClient`.

claude-plugin/commands/status.md
```diff
@@ -0,0 +1,36 @@
...
+1. Detect the operating system.
+
+2. If the devrig launcher is missing, devrig is not installed — tell the user to run
+   `/devrig:setup`, and stop. Check the launcher path for the OS:
+   - **macOS / Linux:** `~/.mcp-steroid/bin/devrig`
+   - **Windows:** `%USERPROFILE%\.mcp-steroid\bin\devrig.cmd`
+
+3. Otherwise run the read-only check with the Bash tool, showing its output:
+   - **macOS / Linux:** `"$HOME/.mcp-steroid/bin/devrig" install claude --check`
+   - **Windows:** `"%USERPROFILE%\.mcp-steroid\bin\devrig.cmd" install claude --check`
```
jonnyzzz, line 23: **major — the Windows branches of `/devrig:status` and `/devrig:uninstall` hand cmd/PowerShell syntax to the Bash tool, which is the only tool these commands allow.**

`allowed-tools` here is `Bash`. Under bash, `%USERPROFILE%` is not expanded and the backslashes are escapes, so this line fails with "no such file or directory" and the user is told devrig is broken when it is fine. Same for the launcher-path check at :19.

`commands/uninstall.md:27` is worse: `Remove-Item -Recurse -Force "$env:USERPROFILE\.mcp-steroid"` is a PowerShell cmdlet given to the Bash tool, so after the user confirms a destructive operation the delete silently never happens.

`commands/setup.md:20` already gets this right: `powershell -NoProfile -ExecutionPolicy Bypass -File "${CLAUDE_PLUGIN_ROOT}/bin/install-devrig.ps1"`.

Fix: use the same `powershell -NoProfile -ExecutionPolicy Bypass -Command "..."` wrapper for the status check and the uninstall delete, and expand `$env:USERPROFILE` inside PowerShell.

claude-plugin/commands/uninstall.md
```diff
@@ -0,0 +1,33 @@
...
+3. **Unregister from Claude** (user scope). Run both, ignoring a "not found" error — the
+   legacy name is cleaned up too:
+   - `claude mcp remove --scope user devrig`
+   - `claude mcp remove --scope user mcp-steroid`
+
+4. **Delete the install directory** with the Bash tool:
+   - **macOS / Linux:** `rm -rf "$HOME/.mcp-steroid"`
+   - **Windows:** `Remove-Item -Recurse -Force "$env:USERPROFILE\.mcp-steroid"`
+
+5. Tell the user devrig has been removed and to **restart Claude**. To reinstall later, they
```
jonnyzzz, line 29: **major — uninstall is not the inverse of install: it leaves the plugin and the marketplace/enabledPlugins entries in place, so the user is left with a permanently failing MCP server and a nag every session.**

After `/devrig:uninstall` deletes `~/.mcp-steroid`:
- the plugin's `.mcp.json` still launches `${CLAUDE_PLUGIN_ROOT}/bin/devrig-mcp`, whose only statement is `exec "$HOME/.mcp-steroid/bin/devrig" mcp` — the binary is gone, so `devrig` shows as failed in `/mcp` forever;
- the `check-devrig` SessionStart hook prints "devrig is not installed yet. Run /devrig:setup" on every single session;
- `enabledPlugins["devrig@jonnyzzz"]` and `extraKnownMarketplaces.jonnyzzz`, written by `ClaudePluginConnect.kt:61-73`, are never removed — and nothing removes them anywhere in the codebase.

Fix: add a final step telling the user to run `/plugin uninstall devrig@jonnyzzz`, and add a `devrig connect claude --disable` inverse (called here) that drops the two settings.json keys for users who arrived via the IDE "Enable" button.

claude-plugin/build.gradle.kts
```diff
@@ -0,0 +1,1137 @@
...
+    group = "verification"
+    description = "Validate .claude-plugin/marketplace.json structure"
+
+    val marketplaceFile = rootProject.projectDir.resolve(".claude-plugin/marketplace.json")
+    // Claude keys a marketplace by the `name` DECLARED here (see ~/.claude/plugins/known_marketplaces.json),
+    // so `devrig connect claude` and the IDE's onboarding check must use the same name — otherwise they
+    // write / look for `devrig@<wrong-marketplace>`, which names no real plugin. Both constants are
+    // cross-checked below.
+    val connectSource = rootProject.projectDir
+        .resolve("npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/ClaudePluginConnect.kt")
```
jonnyzzz, line 639: **major — the build script regex-greps Kotlin *source files* in two other modules to check a constant; it breaks on a legal refactor and has no Gradle dependency behind it.**

`assertConstant` matches `const\s+val\s+$constant\s*=\s*"([^"]*)"`. `npx-kt/.../ClaudePluginConnect.kt:26` already uses the composed form `const val CLAUDE_DEVRIG_PLUGIN_KEY = "devrig@$CLAUDE_MARKETPLACE_NAME"`. Apply that same (correct) dedup to `ij-plugin/.../OnboardingDecision.kt:16` and the regex captures `devrig@`, failing `:claude-plugin:check` with "CLAUDE_DEVRIG_PLUGIN_KEY is 'devrig@' but must be 'devrig@jonnyzzz'". Renaming or moving either file breaks it the same way — from a project whose `plugins { base }` block declares no dependency on either module.

The repo already has the right idiom for scanning production Kotlin: `ij-plugin/src/test/.../NoHardcodedMcpSteroidUriUsageTest.kt`, `NoForbiddenPluginImportsTest.kt`.

Fix: move the invariant into a JUnit test next to the constants — or better, keep one Kotlin constant as the source of truth and *generate* `marketplace.json` from it instead of asserting three hand-written copies agree. (Related: this file is 1137 lines and hosts a behavioural test suite — `validateCheckDevrigRuns` at :444 spawns `sh` against synthetic `$HOME` trees. Those are plain `tasks.register`, not `Test` tasks, so failures give no test report and no per-case granularity. `:installer-gen` already does exactly this job properly with `src/test` + `src/installerIntegrationTest`.)

ij-plugin/src/test/kotlin/com/jonnyzzz/mcpSteroid/onboarding/InstallerProgressTest.kt
```diff
@@ -0,0 +1,73 @@
...
+
+    @Test
+    fun `devrig download is recognised too`() {
+        val step = parseInstallerLine("[mcp-steroid] downloading devrig (~226 MB) from https://example.com/devrig.zip ...")
+        assertEquals(226L * 1024 * 1024, step!!.totalBytes)
+        assertTrue(step.text, step.text.contains("devrig"))
+    }
+
+    @Test
+    fun `retry, verify, reuse, register and ready lines map to phases without a size`() {
```
jonnyzzz, line 33: **major — this file's stated contract is false: it protects against nothing.**

The header (:10-13) says "these are real lines from `install.sh.tmpl` ... If the templates change their phrasing, these tests fail instead of the progress bar silently going blank." But every phrase below is a hardcoded copy of `InstallerProgress.kt`'s own literal, and nothing here reads `install.sh.tmpl` / `install.ps1.tmpl`.

Failure: rename `install.sh.tmpl:77` from `SHA-256 verified: $ia_sha` to `checksum ok: ...` (and the ps1 twin at :121). `parseInstallerLine`'s `body.startsWith("SHA-256 verified")` stops matching, the IDE progress indicator freezes for the rest of a 611 MB download — and this test still passes, because it feeds itself the old literal. Same for `already installed:`, `another install finished first`, `registering devrig`, `devrig binary is ready`, `platform:`, `attempt N/M failed` and the `ERROR: ` prefix. `InstallerGeneratorTest.kt:151-158` pins exactly one of the eight (the `downloading ... (~N MB)` line); the other seven are unguarded in both directions.

Fix: drive it from the rendered scripts — in `InstallerGeneratorTest`, `renderInstallerScripts(...)` and assert each phrase `parseInstallerLine` recognises actually occurs in `scripts.sh` and `scripts.ps` — or move the phrase list into one shared constant both modules read.

ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/onboarding/OnboardingDecision.kt
```diff
@@ -0,0 +1,117 @@
...
+
+/** Locate the `claude` CLI: scan PATH entries, then fall back to ~/.local/bin. Null if not found. */
+fun findClaudeBinary(pathEnv: String?, userHome: Path, windows: Boolean): Path? {
+    val names = if (windows) listOf("claude.exe", "claude.cmd", "claude.bat") else listOf("claude")
+    // Derive the PATH separator from the `windows` parameter — NOT File.pathSeparatorChar (runtime OS),
+    // so the function stays OS-pure and testable for either OS on any host.
+    val pathSeparator = if (windows) ';' else ':'
+    val dirs = buildList {
+        pathEnv?.split(pathSeparator)?.forEach { entry ->
+            if (entry.isNotBlank()) add(Path.of(entry))
```
jonnyzzz, line 89: **minor — `Path.of` on raw PATH entries throws on characters that are legal in a Windows PATH, killing the whole connection-state computation.**

`PATH=C:\Windows;"C:\Program Files\Foo\bin";C:\tools` -> the split yields an entry including the quote characters and `Path.of` throws `InvalidPathException: Illegal char <"> at index 0`. That propagates out of `findClaudeBinary` -> `DevrigConnectionStateService.compute()` -> `refresh()`, where it is swallowed by `catch (e: Exception)` in `DevrigOnboardingService.kt:64` and `DevrigConnectionState.kt:73`.

Result: `cached` stays null for the whole session, the status-bar widget is stuck on "Checking whether this IDE is connected", and the onboarding offer never appears on that machine — with only a log warning.

Fix: `runCatching { Path.of(entry.trim().trim('"')) }.getOrNull()` per entry and skip unparseable ones.

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/server/DevrigWorkspaceRoots.kt
```diff
@@ -0,0 +1,67 @@
...
+    private val log = thisLogger()
+    private val mutex = Mutex()
+
+    @Volatile
+    private var cached: List<Path>? = null
+
+    suspend fun candidateDirs(): List<Path> {
+        cached?.let { return it }
+        return mutex.withLock {
+            cached ?: computeDirs().also { cached = it }
```
jonnyzzz, line 32: **minor — the degraded cwd-only fallback is cached permanently, and `invalidate()` has no production caller.**

If the client's `roots/list` reply exceeds the 3 s budget on the first project-less tool call (`withTimeoutOrNull(3_000)`, `DevrigServices.kt:112`), `computeDirs()` returns null, the code falls back to `listOfNotNull(cwd)` — and stores that in `cached`.

Failure: Claude Code was launched from `$HOME` while the project is `~/work/app`. Every subsequent project-less call in that stdio session reuses the single cwd, `selectProjectByCwd` returns `NoMatch`, and the user gets "No open project matches your working directory ($HOME)" for the rest of the session. `grep -rn 'invalidate()'` finds only the declaration at :37 — nothing calls it, and `notifications/roots/list_changed` is not handled.

Fix: cache only a successful roots result (leave the fallback uncached so the next call retries), and wire `invalidate()` to `roots/list_changed`.

claude-plugin/bin/install-devrig.ps1
```diff
@@ -0,0 +1,49 @@
...
+# devrig_status can report the detail.
+function Write-FailedMarker([string]$Reason) {
+  New-Item -ItemType Directory -Force -Path $MarkerDir | Out-Null
+  Set-Content -Path $Failed -Value $Reason -ErrorAction SilentlyContinue
+  [Console]::Error.WriteLine("devrig: $Reason")
+}
+
+[Console]::Error.WriteLine("devrig: installing devrig via $InstallUrl")
+try {
+  Invoke-RestMethod -Uri $InstallUrl | Invoke-Expression
```
jonnyzzz, line 32: **minor — `exit 1` inside the canonical installer is not a terminating error, so this `catch` never fires and no failure marker is written on Windows.**

`install.ps1.tmpl`'s `Fail` does `Write-Log "ERROR: ..."; exit 1` (tmpl :49), e.g. from the disk-space preflight at tmpl :90-93. `exit` inside `Invoke-Expression` terminates the PowerShell process: the `catch` at :33-36 does not run, `Test-Path $Devrig` at :40 is never reached, and `~/.mcp-steroid/markers/bootstrap-install.failed` is never written.

Next session, `check-devrig` takes the "not installed yet" branch instead of "devrig install failed. Run /devrig:setup to retry" — the user is never told the install actually failed.

Fix: run the canonical installer in a child process (`powershell -NoProfile -ExecutionPolicy Bypass -Command "irm $InstallUrl | iex"`) and branch on `$LASTEXITCODE`.

claude-plugin/bin/devrig-recover
```diff
@@ -0,0 +1,66 @@
...
+:; #   3. indexing / dumb mode                                       -> smartReadAction {} / awaitConfiguration, retry
+:; #   4. modal dialog blocking the EDT                              -> modal=smart_non_modal / closeModalDialogs()
+:; # It stays SILENT ({}) on success and on any UNRECOGNIZED error: the plugin's own ExecutionSuggestionService
+:; # already appends compile/threading tips INSIDE execute_code results, so this hook only fills the gap it
+:; # cannot reach — routing / IDE-state failures, including transport errors that carry no plugin hint at all.
+:; # Stdout is the hook data channel -> emit ONLY JSON. Must exit 0 (a non-blocking PostToolUse hook).
+:; IN=$(cat 2>/dev/null)
+:; emit() { printf '%s\n' "$1"; exit 0; }
+:; # 1. No IDE at all (fundamental — check before project routing, which can present as a side effect).
+:; if printf '%s' "$IN" | grep -qiE 'No IntelliJ IDE with the MCP Steroid plugin is running|no reachable backend|not routable|connection refused|ECONNREFUSED|failed to connect'; then
```
jonnyzzz, line 18: **minor — the signature scan runs on the raw payload with no `isError` check, so repository content can steer the agent after a *successful* call.**

`$IN` is the whole PostToolUse payload, which includes `tool_response` content (repository text returned by `steroid_execute_code`) and `tool_input` (text the model itself wrote). A file containing the literal `connection refused` — a log fixture, a README, test data — matches this grep even when `"isError": false`, and the hook emits `additionalContext`: "RECOVERY: No JetBrains IDE ... is reachable. Do NOT fall back to grep/sed/Bash ... start a managed IDE with the devrig backend CLI".

So repo-controlled bytes can push the agent into starting a managed IDE backend (a multi-hundred-MB download) after a call that succeeded. Same for the other three signatures (`Dumb mode`, `Project not found:`, `modal dialog is present`).

Fix: parse the payload as JSON, run the scan only when `tool_response.isError` is true, and match the error message field rather than the whole payload.

TODO.md
```diff
@@ -1,5 +1,29 @@
...
+  plugin, so the plugin stays bundled inside devrig; devrig fetching the plugin on demand would add a
+  runtime dependency on the plugin repository that can break in our own `backend download/start` flow.
+  Freshness is solved devrig-side instead (devrig keeps itself current → so does the plugin it carries;
+  known work, devrig-owned). Measurements + the full rationale:
+  [`docs/devrig-bundled-in-plugin-spike.md`](docs/devrig-bundled-in-plugin-spike.md).
+- [ ] **The IDE plugin is the migration path onto devrig** — it must move existing plugin users onto a
+  correct, current devrig by running our canonical install scripts. The mechanism already exists
+  (`DevrigSetup.kt:26-28` runs `curl … install.sh | sh` / `irm … install.ps1 | iex`, then
+  `devrig connect claude`). Remaining gaps, in order of value:
+  - **The offer is missable**: the onboarding notification group is `displayType="BALLOON"`
```
jonnyzzz, line 13: **nit — this list is already stale in the same commit, and the line citations point at the wrong lines.**

All five "remaining gaps" are implemented by this PR, and one is deliberately refused by the shipped design:
- `displayType="BALLOON" (plugin.xml:100)` -> `plugin.xml:104-105` already reads `displayType="STICKY_BALLOON"`, and :108-110 registers `DevrigStatusBarWidgetFactory`;
- "Don't ask again" -> explicitly rejected at `DevrigOnboardingService.kt:25-26` ("there is **no 'don't ask again'**");
- ":21 replace `ExecUtil.execAndGetOutput` with streaming" -> done in `runCommandStreaming` (`DevrigSetup.kt:246-250`);
- ":23 analyticsBeacon events" -> present at `DevrigSetup.kt:88` and `:160`;
- ":24 failure marker" -> present at `DevrigSetup.kt:288`.

`OnboardingDecision.kt:36` is cited for `devrigInstalled()`, which is at :42. `DevrigSetup.kt:26-28` is the import block. Same stale text in `docs/devrig-bundled-in-plugin-spike.md:65-79`.

Given the plan to extract small PRs from this branch, a stale gap list is an active hazard: a maintainer reading it will either redo the work or drop `DevrigStatusBarWidget`/`InstallerProgress` as unreferenced. Fix: mark shipped or delete, and reference symbols instead of line numbers — they rot within one commit.

jonnyzzz:
## Review 2026-08-03

- **Superseded by stack #363–#368, per the maintainer plan announced here on 2026-07-16.** The extraction has now happened and the file mapping is 1:1 complete: installer-gen templates/progress/retry → #363; cwd routing (`mcp-steroid-server/CommonToolParams`, `npx-kt/.../server/DevrigProjectDetection|Resolver|RoutingService|WorkspaceRoots` + tests) → #364; `ConnectCommand`/`ConnectIde`/`ClaudePluginConnect`/`InstallCommand`/`Cli` → #365; the entire `claude-plugin/` module + `.claude-plugin/marketplace.json` + root `build.gradle.kts`/`settings.gradle.kts` → #366; `ij-plugin/onboarding/*`, `ServerUrlWriter`, `UpdateChecker`, `plugin.xml`, `docs/devrig-bundled-in-plugin-spike.md` → #367; `release/release-instructions.md`/README → #368. No file in this PR lacks a stack counterpart.
- **Branch is unmergeable and deeply stale**: mergeable=CONFLICTING, 117 ahead / 299 behind main. Merged work semantically conflicts with several hunks: #399 locked the install-scripts contract (install scripts do only install+PATH — this PR's `install.sh.tmpl`/`install.ps1.tmpl` and `claude-plugin/bin/install-devrig` predate it), #397 centralized agent-CLI error handling, #395/#394 reworked `plugin.xml` description + settings, and the #380-series auto-update supersedes the `UpdateChecker.kt` edits here.
- The 2026-07-28 quorum blockers (notably `InstallCommand.kt:15` server rename `mcp-steroid` → `devrig` breaking existing `mcp__mcp-steroid__*` allowlists) remain unaddressed on this branch; they must be verified as fixed in the stack extraction, not here.
- Linked issues: #134 and #160 already closed; #137 (auto-install via setup command) and #201 (hot-swap MCP proxy) are still open and should be re-targeted to the stack — merging #140 as-is would not be the vehicle for them anyway.
- Per the 2026-07-16 decision this PR stays open as the authorship/reference branch until the #308 ledger accounts for the extracted work; recommend closing it (with the ledger link) once #366 lands on main.

**Verdict: superseded** — stack #363–#368 (specifically #366 for the Claude plugin module) delivers this PR's entire scope from a clean base; do not rebase or merge this branch.

Issue #91, opened 2026-06-11
Epic: agent-experience gaps from the multi-agent dogfooding experiment (29 validated items, 2026-06)


jonnyzzz:
## Epic: agent-experience gaps found by the multi-agent dogfooding experiment (2026-06)

**Research document (full evidence per item):** [docs/agent-feature-review-2026-06.md](https://github.com/jonnyzzz/mcp-steroid/blob/main/docs/agent-feature-review-2026-06.md) ([permalink @ bef9a015](https://github.com/jonnyzzz/mcp-steroid/blob/bef9a015a97ce5df05c012a5a74781b4c281424d/docs/agent-feature-review-2026-06.md))

**Method.** Two independent agent tracks — a 4-lens Claude swarm and Codex (via run-agent.sh) — explored the **live** tooling hands-on (steroid_* MCP tools against running IDEA/GoLand/idea-community backends + the devrig CLI; codex alone made 22 `steroid_execute_code` calls) and each produced a top-20 with reproduced evidence. Merged to a ranked top-30, then **5 validation iterations** in which agents re-tested every claim against the live tooling (refuted claims dropped, new finds admitted), then a final **3-lens quorum** (agent value / philosophy fit / evidence rigor) with hands-on spot re-reproduction. Every item carries a concrete reproduced citation (`eid_*` execution ids) at HEAD `bb738d57`. One top-30 entry was refuted and dropped (refactoring dry-run guidance — already shipped); ranks 1–26 are quorum-validated, 27–29 contested (2-of-3 approval).

Per `docs/PHILOSOPHY.md`, most items are **recipes / response-shape / reliability** work — not new tools.

### Theme A — verification loop (edit → verify is broken in places)
- [ ] **1.** Canonical "verify no red code after edit" recipe that works unfocused (`recipe`, validated)
- [x] **2.** Single-call "run test class → structured results" recipe + SMTRunner `getDelegate()` unwrap for 2026.1 (`recipe`, validated)
- [x] **5.** Verification primitives must distinguish "clean" from "did not run" (`response-shape`, validated)
- [ ] **6.** Structured build-error extraction for ProjectTaskManager/JPS builds (`recipe`, validated)
- [ ] **19.** Module-scope inspection run recipe (GlobalInspectionContext + AnalysisScope, 3 pitfalls documented) (`recipe`, validated)

### Theme B — execute_code response shape
[...]

PR #142, opened 2026-06-23
Theme A verification primitives: distinguish clean from did-not-run

matveidetonator:
Completes Theme A (issue #91): every verification primitive must distinguish clean from did-not-run.

Inspection primitive: runInspectionsDirectly now reports a non-PSI VirtualFile (e.g. a directory) via failedTools (sweep-failure id) instead of an empty map that looks clean; input/setup failures log at warn, real crashes stay errors. Map-compatible surface unchanged.

New recipes :
- ide/verify-after-edit: canonical result vocabulary (clean | findings | errors | passed | failed | did_not_run | check_failed); empty diagnostics is never success without check_ran=true.
- ide/jps-build-errors: structured JPS diagnostics via public CompilerManager.make / CompileContext / CompilerMessage on the EDT; make always runs JPS regardless of Gradle/Maven delegation, but results can diverge from the external build.
- ide/module-inspection-sweep: module-scope loop over runInspectionsDirectly, avoiding internal GlobalInspectionContextImpl and deprecated createNewGlobalContext.
- test/run-test-class-structured: one-call structured Gradle test run; correlates TEST_STATUS events by public SMRootTestProxy.getHandler identity (no console cast, no internal BuildView).

Tests: VerificationRecipeStatusContractTest, RunTestClassStructuredRecipeContractTest, runInspectionsDirectly non-PSI case; fixed the pre-existing red inspect-and-fix example test (RedundantCast). Also removed redundant JSON-RPC id assertions in McpServerCore.


prompts/src/main/prompts/test/run-test-class-structured.md
```diff
@@ -0,0 +1,343 @@
...
+                "testPattern" to testPattern,
+                "gradleTask" to gradleTestTaskPath,
+                "message" to "RunContentDescriptor has no process handler.",
+            )
+        )
+        return
+    }
+
+    targetProcessHandler.set(handler)
+    synchronized(startedRoots) {
```
jonnyzzz: synchronized? 

it probably looks like a mistake or misuse of the API, since the code is in a suspend function already it can leverage something more better either from coroutines or JVM

TODO.md
```diff
@@ -1,5 +1,11 @@
 # TODO
 
+- [ ] **runInspectionsDirectly follow-ups (#69 ask 1, deferred)**: a `PsiFile`-accepting overload
```
jonnyzzz: incomplete?

prompts/src/main/prompts/test/run-test-class-structured.md
```diff
@@ -0,0 +1,343 @@
+Test: Run Gradle Test Class and Return Structured Results
```
jonnyzzz, line 1: this is about test runs, nice to have it in the next/prev PR

PR #163, opened 2026-06-27
test-experiments: provide test.integration.devrig.package.zip (unblocks all experiment builds)

jonnyzzz:
Every :test-experiments:test (arena/youtrackdb/SSR/Android Studio) was failing at init — the infra requires `test.integration.devrig.package.zip` and :test-experiments didn't set it (only :test-integration did). Mirrors :test-integration's devrigPackageDist wiring.

### DIFF
--- a/test-experiments/build.gradle.kts
+++ b/test-experiments/build.gradle.kts
@@ -21,9 +21,22 @@ val agentOutputFilterDist by configurations.creating {
     isCanBeResolved = true
 }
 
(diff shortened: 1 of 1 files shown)


jonnyzzz:
Redundant — origin/main already contains the devrig.package.zip wiring. The fix that mattered (JetBrains/mcp-steroid main, which the TC builds check out) landed via JetBrains#4.

PR #170, opened 2026-06-29
fix(exec): waitForSmartMode timeout = "indexing in progress, keep polling" (#169)

### LINKED ISSUE #169: YouTrackDb experiments fail on v101: waitForSmartMode 60s cap too short for large projects (regression vs v93)
## Summary

All four **YouTrackDb** test-experiment builds fail on plugin **v101 (0.101.572)**, while the identical scenarios **passed on v93 (0.93.0.524)**. The failure is a **regression in the plugin**, not in the test or the IDE: the `waitForSmartMode()` 60-second deadlock guard (added 2026-05-31 in `fe6d9391`) is too short for a very large project, so the test-harness project-setup step times out before YouTrackDb finishes indexing.

## Affected builds (TeamCity, all on 0.101.572-jb-5098ae5)

| Build | Config | Result |
|-------|--------|--------|
| #987034136 | YouTrackDb_Default_Claude   | FAILURE (tests: 1 failed) |
| #987034138 | YouTrackDb_Default_Codex    | FAILURE (tests: 1 failed) |
| #987034140 | YouTrackDb_Pinned261_Claude | FAILURE (tests: 1 failed) |
| #987034142 | YouTrackDb_Pinned261_Codex  | FAILURE (tests: 1 failed) |

All four fail identically with `YouTrackDbMavenTest.initializationError` (the `@BeforeAll` setup, not the agent run).

## Evidence (from the build logs)

```
[IntelliJ LOG] SEVERE - McpScriptContextImpl - [eid_...-integration-test]
  waitForSmartMode did not reach smart mode within 1m
[IDE OUT] steroid_execute_code result, isError=true:
  [PRE] close modal dialogs
  [PRE] sync documents
  [PRE] wait for smart mode
  Waiting for indexing to complete...
  FAILED: waitForSmartMode did not reach smart mode within 1m — indexing may be stuck.
→ com.jonnyzzz.mcpSteroid.integration.tests.YouTrackDbMavenTest.initializationError
```

The execution id ends in `-integration-test` (the driver's default `task_id`), so this is the **test harness's own setup call** — `waitForProjectReady(MAVEN)` → `mcpExecuteCode(...)` — whose default `modal=DEFAULT` runs a `[PRE] waitForSmartMode` before the script body. It is **not** the AI agent's call.

## Root cause — confirmed regression

`ij-plugin/.../execution/McpScriptContextImpl.kt`:

```kotlin
/** Deadlock guard for [waitForSmartMode] when indexing never reaches smart mode. */
private val WAIT_FOR_SMART_MODE_TIMEOUT = 60.seconds
```

- Before commit `fe6d9391` (2026-05-31), `waitForSmartMode()` was an **unbounded** `suspendCancellableCoroutine` — it waited as long as indexing needed.
- `fe6d9391` wrapped it in `withTimeout(60.seconds)` as a deadlock guard.
- v93 (build dated 2026-05-01) predates that commit → unbounded wait → **PASS**.
- v101 (build dated 2026-06-29) has the 60s cap → **FAIL** on YouTrackDb.

The IDE is **not** the variable: `Pinned261` uses the same IDE (2026.1.1) in both the passing v93 and the failing v101 run. Only the plugin version changed.

YouTrackDb (`https://github.com/JetBrains/youtrackdb`) is a large multi-module Maven project — `YouTrackDbMavenTest`'s own doc calls it "many submodules — picks heavy work for any agent." Its first post-import indexing legitimately takes longer than 60s on the CI agent, so the very first driver `steroid_execute_code` setup call's `[PRE] waitForSmartMode` times out.

## Why this matters beyond YouTrackDb

The 60s cap is a hard ceiling on *any* `steroid_execute_code` call's PRE sweep. Real agents hitting a large repo right after import will see the same `waitForSmartMode did not reach smart mode within 1m` error even when indexing is simply *slow*, not *stuck*. A "deadlock guard" should distinguish minutes-of-legitimate-indexing from a true hang.

## Suggested fixes (pick one)

1. **Raise the guard** to a value that comfortably exceeds large-project indexing (e.g. 5–10 min). A deadlock that never resolves is caught just as well by 10m as by 1m; slow indexing is not.
2. **Make it configurable** via a Registry key / system property so heavy-project tests (and large real repos) can extend it without a code change.
3. **Scale to project size** or reset the timer on indexing progress (timeout on *no progress* for 60s, not on *total* time).

Option 1 or 2 is the minimal fix to make YouTrackDb green again on v101.

## Repro

`./gradlew :test-experiments:test --tests '*YouTrackDbMavenTest*'` against a fresh checkout of `JetBrains/youtrackdb`, on a CI-class agent where indexing exceeds 60s.

jonnyzzz:
Fixes #169. **Third revision — minimal, per review.** Earlier revisions (raise+configure the timeout via a Registry key; a `RetryableToolCallException`/`isRetryable` mechanism) were both dropped as over-engineered.

## The point

`waitForSmartMode()` (the `[PRE] wait for smart mode` step of every default-modal `steroid_execute_code`) treated its 60s timeout as a hard failure — *"did not reach smart mode within 1m — indexing may be stuck"*, plus a `SEVERE` log and a thread dump. On a large project still doing its first index, that's wrong on every count: nothing is stuck — indexing is just still running and **always making progress**. It broke all four YouTrackDb test-experiments on v101 (~105k files; setup's first call, `mcpRegisterJdks`, died in its PRE sweep before the Maven import even ran). Evidence + thread dump in #169.

Indexing-not-done is never a failure, and the right response is always the same — wait and try again. So **the fix is the message, not machinery.**

## Change (2 files + 1 test)

- **`waitForSmartMode`** keeps the short 60s per-call window, but on timeout returns a clear, friendly result: *"INDEXING IN PROGRESS … this is normal and expected, not an error and nothing is stuck; indexing always makes progress. A medium-size project can take up to an hour, larger ones longer — that wait is acceptable. Keep polling: call this tool again to continue waiting."* Dropped the misleading "may be stuck", the `SEVERE` log, and the per-poll thread dump.
- **Test harness `mcpExecuteCode`** polls on that message — exactly as an agent would — bounded by a generous overall budget, so a slow first index doesn't break setup.

No new exception types, no `isError`/retryable flags, no registry keys, no change to the 60s window.

## Test

`test-integration` `IndexingInProgressDetectionTest` covers the poll predicate (`isIndexingInProgress`): detects the message, ignores success and genuine failures. Plugin compiles; existing execution suites green.

## Reproduction note

The CI failure is environment-bound to slow CI disk: locally a fast SSD finishes indexing under 60s, so the message/poll path never triggers (the local run sails through setup and only fails later for a missing agent API key). The poll path is covered by the unit test and will be exercised by the YouTrackDb builds on CI.

### DIFF
--- a/ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/execution/McpScriptContextImpl.kt
+++ b/ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/execution/McpScriptContextImpl.kt
@@ -201,8 +201,9 @@ class McpScriptContextImpl(
         resultBuilder.logProgress("Waiting for indexing to complete...")
 
         try {
-            // Bounded as a deadlock safety net (indexing that never reaches smart mode) — the tool docs
-            // promise the wait is bounded. Timeout => fail the execution with a clear message.
+            // We only wait this short window per call so the request returns promptly. Reaching it does
+            // NOT mean anything is wrong: dumb mode means indexing is actively running and making
+            // progress. We tell the caller so, and let it poll (call again).
             withTimeout(WAIT_FOR_SMART_MODE_TIMEOUT) {
                 suspendCancellableCoroutine<Unit> { cont ->
                     fun waitForSmart() {
@@ -224,11 +225,15 @@ class McpScriptContextImpl(
                 }
             }
         } catch (_: TimeoutCancellationException) {
-            captureThreadDump("waitForSmartMode-timeout")
-            log.error("[$executionId] waitForSmartMode did not reach smart mode within $WAIT_FOR_SMART_MODE_TIMEOUT")
+            // Not an error and not stuck — indexing is in progress. Tell the caller to keep polling.
+            log.info("[$executionId] still indexing after $WAIT_FOR_SMART_MODE_TIMEOUT — asking caller to poll")
             throw ToolCallErrorException(
-                "waitForSmartMode did not reach smart mode within $WAIT_FOR_SMART_MODE_TIMEOUT — indexing may " +
-                    "be stuck. See the thread dump under execution '${executionId.executionId}'."
+                "INDEXING IN PROGRESS: the IDE is still indexing this project, so it is not ready yet. " +
+                    "This is normal and expected — it is NOT an error and nothing is stuck; indexing always " +
+                    "makes progress and simply needs more time. Indexing a medium-size project can take up " +
+                    "to an hour, and larger projects longer; that wait is acceptable. Just keep polling: " +
+                    "call this tool again to continue waiting (each call waits up to $WAIT_FOR_SMART_MODE_TIMEOUT), " +
+                    "and your request runs automatically once indexing finishes."
             )
         } finally {
             log.info("[$executionId] Waiting for indexing completed")
--- a/test-integration/src/main/kotlin/com/jonnyzzz/mcpSteroid/integration/infra/mcp-steroid.kt
+++ b/test-integration/src/main/kotlin/com/jonnyzzz/mcpSteroid/integration/infra/mcp-steroid.kt
@@ -66,6 +66,19 @@ internal fun ProcessResult.resolveJavaHomeLookup(jdkVersion: String): String {
     error("[COMPILE] JDK $jdkVersion lookup returned no path; stdout=${stdout.take(500)} stderr=${stderr.take(500)}")
 }
 
+/**
+ * Marker the server's `waitForSmartMode` emits when the IDE is still indexing — the signal to keep
+ * polling (call again). Mirrors the message in `McpScriptContextImpl.waitForSmartMode`; the coupling is
+ * the documented tool-result contract.
+ */
+internal const val INDEXING_IN_PROGRESS_MARKER = "INDEXING IN PROGRESS"
+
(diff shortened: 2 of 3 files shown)


jonnyzzz:
Local sanity re-run against the **new** plugin (`:test-experiments:test -PtestFilter='*YouTrackDbMavenTest*claude*'`): every setup `[PRE] wait for smart mode` step completed with `isError:false` (JDK reg → Maven import ×"Smart mode reached" → SDK resolve → plugin install → UX), **no fatal timeout, no RETRYABLE poll triggered** (local SSD indexes <60s), and the run fails only later at the expected `Claude Code API key not found`. Happy path intact — no regression. The retry/poll path itself can't be exercised on fast local disk; it's covered by the unit tests and will be validated by the YouTrackDb builds on CI.

jonnyzzz:
Reworked to the minimal approach: the fix is just the `waitForSmartMode` timeout **message** — it now says indexing is in progress, always makes progress, and that long waits (up to ~1h for medium projects, longer for larger) are acceptable, so the caller should keep polling. Dropped the `SEVERE`/thread-dump and the earlier `RetryableToolCallException`/`isRetryable` machinery entirely. The harness just polls on that message like an agent would. Force-pushed.

PR #178, opened 2026-06-29
feat(test-helper): route containerized agents through a configurable LLM gateway

jonnyzzz:
Lets the test agents (Claude, Codex, Gemini) route their LLM traffic through a host-configured OpenAI/Anthropic-compatible **gateway** instead of the public vendor API. No-op on CI (nothing set → agents use their key against the vendor API, as today).

**How**
- When a per-vendor base URL is set on the host, forward it into the agent container, rewriting a loopback host (`127.0.0.1`/`localhost`/`0.0.0.0`) → `host.docker.internal` (the docker-start helper adds that alias). Single source of truth: the shared `DOCKER_HOST_ALIAS` constant.
- **Claude** (`ANTHROPIC_BASE_URL`) and **Gemini** (`GOOGLE_GEMINI_BASE_URL`) honor the env var directly.
- **Codex** ignores `OPENAI_BASE_URL` (verified — it connects to `api.openai.com` regardless), so it only routes via a *configured provider*. When a gateway URL is present we point Codex at it with **`-c model_provider=…` override flags derived from the same env var — no config file written**.

**Validated locally** through a host gateway: `AgentEndpointTest` (rewrite) green; **Claude + Codex both reach the gateway end-to-end** (no `api.openai.com`, no 401). Pure, vendor-neutral — no third-party CLI or proxy named anywhere (squashed to one clean commit).

### DIFF
--- a/test-helper/src/main/kotlin/com/jonnyzzz/mcpSteroid/testHelper/AgentEndpoint.kt
+++ b/test-helper/src/main/kotlin/com/jonnyzzz/mcpSteroid/testHelper/AgentEndpoint.kt
@@ -1,0 +1,29 @@
+/* Copyright 2025-2026 Eugene Petrenko (mcp@jonnyzzz.com); Copyright 2025-2026 JetBrains. Use of this source code is governed by the Apache 2.0 license. */
+package com.jonnyzzz.mcpSteroid.testHelper
+
+import com.jonnyzzz.mcpSteroid.testHelper.docker.DOCKER_HOST_ALIAS
+
+/**
+ * Optional per-vendor API base URL for the containerized agents.
+ *
+ * The agent CLIs (Claude Code, Codex, …) honor the standard `<VENDOR>_BASE_URL` environment variables to
+ * route their LLM traffic through an OpenAI/Anthropic-compatible endpoint instead of the public API. When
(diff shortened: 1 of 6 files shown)


jonnyzzz:
**Why Codex uses `-c` provider flags instead of just an env var** (Claude/Gemini use the env var directly):

The Codex CLI has **no base-URL environment variable** — it reaches the cloud model **only** through a configured provider (`model_provider` + `[model_providers.*].base_url`), and ignores `OPENAI_BASE_URL` entirely. Verified empirically: with only `OPENAI_BASE_URL=<gateway>` set, Codex still connects to `https://api.openai.com/v1/responses` (the public API) and 401s with the gateway key. `codex --help` confirms there's no `*_BASE_URL` override for the cloud provider.

So redirecting Codex to a gateway **requires** specifying a provider. We do that the lightest way — Codex's own `-c model_provider=… -c model_providers.gateway.base_url=…` command-line overrides, **no config file written** — and the URL is derived from the **same env var** the other agents use; auth still flows through `OPENAI_API_KEY`. So it's env-sourced, just applied the only way Codex accepts.

This rationale is now in the code comment in `DockerCodexSession.runPrompt`. Validated locally: Claude + Codex both route through the gateway (no `api.openai.com`, no 401).

PR #181, opened 2026-06-30
test(build): add class-file version guard for ij-plugin + npx-kt

jonnyzzz:
## Why

The plugin and devrig are compiled with a JDK 25 toolchain, emitting **class-file major 69**. They load in IntelliJ IDEA 2026.1 (bundled **JBR 25**) but **not** in Android Studio on the same 261 platform, which bundles **JBR 21** (class-file major 65) and throws `UnsupportedClassVersionError`. Nothing in the build caught this.

## What

`VerifyClassFileVersionTask` (buildSrc) reads each `.class` header directly (no ASM) and fails the build if any class an artifact ships exceeds the maximum Java feature the **oldest supported JBR** allows.

- **Recursive over every bundled jar *and* zip at any folder depth** — the plugin's `kotlinc/` and `ocr-tesseract/` trees, and the `ij-plugin.zip` that the devrig dist bundles — not just a `lib/` directory.
- **Multi-release-JAR aware** — `META-INF/versions/<N>/` overlays for `N` newer than the floor are skipped (they never load on that runtime); base + at/below-floor overlays are checked.
- Skips `module-info.class` (a module descriptor, not a classpath-loaded class).
- Entries named `*.class` that don't parse (bad magic / truncated) are reported as **warnings** instead of being silently dropped.
- Pure scan logic in `ClassFileVersionScanner`, covered by buildSrc unit tests (header parsing incl. bad-magic/short, MRJAR floor rule, module-info skip, broken-class reporting, recursive nested jar/zip descent).
- Wired into **`:ij-plugin`** (plugin zip) and **`:npx-kt`** (devrig dist) via the same lifecycle hooks as `verifyBundledLibraries` (`test` / `buildPlugin`+`distZip` finalizer / `verifyPlugin` / `check`).

`maxJavaFeature` is **25** here, locking the current state. It is the single knob a follow-up lowers to **21** alongside the JDK-target change.

## Verification (local)

- buildSrc unit tests: `tests=4, failures=0`.
- Recursive guard scans real artifacts green, no broken-class warnings: `:ij-plugin` **62,577** classes, `:npx-kt` **82,367** classes ≤ major 69.

First of a 3-part series: (1) **this guard**, (2) JDK 21 target (`-jvm-target 21` + `-Xjdk-release=21`) + lower the knob to 21, (3) Android Studio integration test.

### DIFF
--- a/buildSrc/src/main/kotlin/com/jonnyzzz/mcpSteroid/gradle/ClassFileVersionScanner.kt
+++ b/buildSrc/src/main/kotlin/com/jonnyzzz/mcpSteroid/gradle/ClassFileVersionScanner.kt
@@ -1,0 +1,128 @@
+/* Copyright 2025-2026 Eugene Petrenko (mcp@jonnyzzz.com); Copyright 2025-2026 JetBrains. Use of this source code is governed by the Apache 2.0 license. */
+package com.jonnyzzz.mcpSteroid.gradle
+
+import java.io.ByteArrayInputStream
+import java.io.File
+import java.io.InputStream
+import java.util.zip.ZipFile
+import java.util.zip.ZipInputStream
+
+/**
+ * Pure, Gradle-independent class-file version scanning used by [VerifyClassFileVersionTask]. Reads the
+ * `.class` header bytes directly — no ASM / bytecode library — so the multi-release-JAR and header-parsing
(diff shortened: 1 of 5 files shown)


jonnyzzz:

PR #184, opened 2026-06-30
test(integration): Android Studio runtime-compat gate with exec_code

jonnyzzz:
Part 3 of 3, builds on #181 (guard) and #182 (JVM 21 target).

## What

**1. Android Studio runtime-compat test** — `AndroidStudioRuntimeCompatTest` installs the production plugin into **Android Studio 261** (stable "Quail" — same IntelliJ 261 platform as IDEA 2026.1, but bundles **JBR 21** where IDEA bundles JBR 25) and exercises the core MCP tools: `list_projects`, `list_windows`, and an **`execute_code`** call.

**2. Fast-fail plugin-load detection** — `waitForMcpReady` uses a small poll helper (`waitUntilReady` + `PollOutcome`) that aborts the moment a load failure appears, instead of spinning the health check until the 300s deadline. The detector keys on `java.lang.UnsupportedClassVersionError` in `idea.log` — **version-agnostic, no hardcoded class-file version** — so it stays correct as bundled JBRs change. `findUnsupportedClassVersionError` and `waitUntilReady` are pure and **unit-tested** (`IdeTestHelpersTest`).

**3. Cleaner guard failure message** — `verifyClassFileVersions` no longer prints class-file version numbers; it names the offending classes and the Java target only.

## Host requirement

Android Studio publishes only a Linux **x86_64** archive (no Linux arm64), and the IDE container runs the host architecture — so this test runs on x86_64 CI and **cannot** run on an Apple-Silicon dev box (fails fast at download resolution there). The architecture-independent regression is covered locally on any host by the bytecode guard.

## Verification (local, arm64)

- `:test-integration:test --tests '*IdeTestHelpersTest*'` → `tests=13, failures=0` (incl. the new detector + poll-helper tests).
- Guards green: `:ij-plugin` / `:npx-kt` classes compiled for Java 21 or older.
- **Not run locally:** the Android Studio container itself (arm64 host — see above); it executes on the x86_64 CI lane.

### DIFF
--- a/buildSrc/src/main/kotlin/com/jonnyzzz/mcpSteroid/gradle/VerifyClassFileVersionTask.kt
+++ b/buildSrc/src/main/kotlin/com/jonnyzzz/mcpSteroid/gradle/VerifyClassFileVersionTask.kt
@@ -58,22 +58,20 @@ abstract class VerifyClassFileVersionTask : DefaultTask() {
         require(checked > 0) { "Scanned 0 class files across ${archives.files} — the guard verified nothing." }
 
         if (violations.isNotEmpty()) {
-            val maxMajor = maxFeature + ClassFileVersionScanner.CLASS_MAJOR_OFFSET
             throw GradleException(
                 buildString {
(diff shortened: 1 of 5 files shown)


jonnyzzz:

PR #185, opened 2026-06-30
test-integration: always download Maven sources/docs; simplify import polling to waitForValue

jonnyzzz:
Supersedes this PR's original "source-download off + bespoke poll" approach, per review.

## What

- **Always auto-download library sources + javadoc** for every project and test — agents get full API docs in the editor, a deliberate must-have. Removes the `downloadSourcesAndDocs` param and the per-project `downloadMavenSourcesAndDocs` flag (including Keycloak's `false` override).

- **Drop `pollWhileIdeBusy`** (custom 3s-interval loop with injected `now`/`sleep`) and reuse the existing **`waitForValue`** (50 ms). `mcpExecuteCode` retries the steroid call while the IDE reports busy:
  - a clean **`INDEXING IN PROGRESS`** result, or
  - a **transient killed request** (exit −1, IDE too saturated to answer)
  → `null` → `waitForValue` calls again. A non-transient failure aborts the wait immediately via `WaitAbortedException` (now carries a `cause`), so a real bug is never masked by an hour of polling.

- Replace `PollWhileIdeBusyTest` with `McpBusyRetryTest` (the two pure retry predicates `isIndexingInProgress` / `isTransientMcpRequestFailure`); `MavenImportSourceDownloadTest` now asserts source/doc download stays on.

Net **−178 lines**.

## Tradeoff

Always-downloading sources reverses the original Keycloak "settle on CI" tactic (which turned source-download *off*). Keycloak's heavy Maven import now relies entirely on the `waitForValue` retry-through-saturation path instead. Deliberate: the in-editor API docs are worth it.

## Verification (local)

- `McpBusyRetryTest` 2/2, `MavenImportSourceDownloadTest` 2/2; `:test-integration` compiles.


test-integration/src/main/kotlin/com/jonnyzzz/mcpSteroid/integration/infra/IdeTestHelpers.kt
```diff
@@ -112,6 +113,34 @@ object IdeTestFolders {
...
     val ideDownloadDir: File = File(System.getProperty("test.integration.ide.download.dir", "build/ide-download"))
 
+    /**
+     * Host directory persisting the container's Maven (`~/.m2`) and Gradle (`~/.gradle`) caches across
+     * runs, so library jars + their sources/javadoc (always auto-downloaded on import) are resolved once
+     * and reused — turning a multi-minute cold Keycloak import into a warm one. Set via
+     * `test.integration.dependency.cache.dir` (the Gradle build points it at a root-shared dir so the
+     * integration and experiments suites — which never run concurrently — share one cache).
+     */
+    val dependencyCacheDir: File =
```
jonnyzzz: assert is missing -- add it

test-integration/src/main/kotlin/com/jonnyzzz/mcpSteroid/integration/infra/IdeTestHelpers.kt
```diff
@@ -112,6 +113,34 @@ object IdeTestFolders {
...
+        File(System.getProperty("test.integration.dependency.cache.dir", "build/test-dependency-cache"))
+
+    /**
+     * Bind mounts that persist the container's `~/.m2` and `~/.gradle` to host dirs under
+     * [dependencyCacheDir]. The host dirs are made world-writable: Linux bind mounts do not remap UIDs, so
+     * the in-container `agent` (uid 1000) must be able to write them (see [allocRunDirAndTitle]). `agent`'s
+     * home is `/home/agent` (see the ide-base Dockerfile).
+     */
+    fun dependencyCacheVolumes(): List<ContainerVolume> =
```
jonnyzzz: dedicated folder per project is ok, just simplify

test-integration/src/main/kotlin/com/jonnyzzz/mcpSteroid/integration/infra/mcp-steroid-import.kt
```diff
@@ -32,6 +38,11 @@ fun McpSteroidDriver.mcpTriggerImportAndWait(buildSystem: BuildSystem) {
                 }
             """.trimIndent()
 
+fun McpSteroidDriver.mcpTriggerImportAndWait(buildSystem: BuildSystem) {
+    //TODO: move that to prompts and include it from there are resources
+    val triggerCode = when (buildSystem) {
+        BuildSystem.MAVEN -> mavenImportTriggerCode()
```
jonnyzzz, line 44: unify with Gradle logic

test-integration/src/main/kotlin/com/jonnyzzz/mcpSteroid/integration/infra/mcp-steroid-import.kt
```diff
@@ -110,27 +121,43 @@ fun McpSteroidDriver.mcpTriggerImportAndWait(buildSystem: BuildSystem) {
             """.trimIndent()
     }
 
+    // Cold imports with always-on source/javadoc download have a long indexing tail: each downloaded
+    // *-sources.jar adds a library root and re-triggers indexing, so the project bounces in and out of
+    // dumb mode in waves. The host-mounted ~/.m2 cache keeps re-runs fast.
+    val settleTimeoutMs = 20 * 60 * 1000L
```
jonnyzzz, line 127: instead check if progress is moving and it is not stuck

test-integration/src/main/kotlin/com/jonnyzzz/mcpSteroid/integration/infra/mcp-steroid.kt
```diff
@@ -91,6 +91,24 @@ internal const val MCP_SERVER_STARTUP_FAILURE_MARKER = "Failed to start MCP serv
...
+ * This is the silent twin of [isIndexingInProgress]: when the IDE settles enough to answer, it returns
+ * the clean `INDEXING IN PROGRESS` marker and we poll; but while the IDE is *so* busy it can't even
+ * answer in time, the request is killed and surfaces as `assertExitCode` throwing with an `exit code is
+ * -1` message (see [executeMcpRequestRaw] -> `assertExitCode(0)`). Both mean the same thing — "still
+ * busy, call again" — so both must be retried. A *script-level* error (compile/runtime) instead comes
+ * back as a clean isError result with a real exit code (1), never as this killed-process -1, so it is
+ * NOT transient and must surface immediately. See jonnyzzz/mcp-steroid#169.
+ */
+internal fun isTransientMcpRequestFailure(t: Throwable): Boolean {
+    val msg = t.message ?: return false
```
jonnyzzz: this one can be wrapped inside, so the McpRequestFailure throws the error that stops wait for, so no need for thaat funnction, it will be implicit for all cases

test-integration/src/main/kotlin/com/jonnyzzz/mcpSteroid/integration/infra/mcp-steroid.kt
```diff
@@ -411,17 +429,23 @@ try {
...
+        //   2. a transient transport failure — the request was killed (exit -1) because the IDE couldn't
+        //      even answer in time (#169).
+        // A non-transient failure (a real bug) instead aborts the wait at once via WaitAbortedException,
+        // so it is never masked by an hour of polling. waitForValue retries on null until the budget runs out.
+        waitForValue(INDEXING_POLL_BUDGET_MS, "exec_code '$taskId' to run (IDE busy with import/indexing)") {
+            val result = try {
+                mcpExecuteCodeOnce(code, taskId, reason, timeout, projectName, modal)
+            } catch (t: Throwable) {
+                if (isTransientMcpRequestFailure(t)) return@waitForValue null // still busy — call again
+                throw WaitAbortedException("exec_code '$taskId' failed: ${t.message}", t)
```
jonnyzzz: make mcpExecuteCodeOnce manage and throw correct exception automatically to break the waitFor loop

test-integration/src/test/kotlin/com/jonnyzzz/mcpSteroid/integration/infra/MavenImportSourceDownloadTest.kt
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Pure tests for [mavenImportTriggerCode] — the script `mcpTriggerImportAndWait` runs to trigger the
+ * Maven import. Library sources + javadoc auto-download is a deliberate must-have (agents get full API
+ * docs in the editor), so the generated script must always enable it.
+ */
+class MavenImportSourceDownloadTest {
+
+    @Test
+    fun `always enables source and doc auto-download`() {
```
jonnyzzz, line 15: expand that to Gradle too, there must be the same experience

test-integration/build.gradle.kts
```diff
@@ -114,6 +114,13 @@ fun Test.configureIntegrationTest(sourceSetName: String = "test") {
             "test.integration.repo.cache.dir",
             layout.buildDirectory.dir("repo-cache").get().asFile.absolutePath,
         )
+        // Persisted container Maven (~/.m2) + Gradle (~/.gradle) caches, shared (root build dir) with
+        // :test-experiments — the two suites never run concurrently — so deps + sources download once and
+        // are reused across runs. See IdeTestFolders.dependencyCacheVolumes.
+        systemProperty(
+            "test.integration.dependency.cache.dir",
+            rootProject.layout.buildDirectory.dir("test-dependency-cache").get().asFile.absolutePath,
```
jonnyzzz, line 124: LGTM, we also have a cache of downloaded IDEs, does it make sesnse to align?

test-integration/src/main/kotlin/com/jonnyzzz/mcpSteroid/integration/infra/IdeTestHelpers.kt
jonnyzzz: Done in a4b9c07f — the soft default is gone: `dependencyCacheDir` now fails fast via `error(...)` when the property is unset, matching `repoCacheDir` (`ideDownloadDir` got the same assert for one consistent convention).

test-integration/src/main/kotlin/com/jonnyzzz/mcpSteroid/integration/infra/IdeTestHelpers.kt
jonnyzzz: Done in a4b9c07f — the `mapOf(...).map { (guest, host) -> ... }` construction is replaced by a flat `listOf(dependencyCacheVolume("m2", ...), dependencyCacheVolume("gradle", ...))` with one small private helper; behavior identical.

test-integration/src/main/kotlin/com/jonnyzzz/mcpSteroid/integration/infra/mcp-steroid-import.kt
jonnyzzz: Done in #199 — Maven now subscribes to the project-level `MavenImportListener.TOPIC` before the trigger and awaits `importFinished` (8-min bound, no blind delay), mirroring Gradle's `ProjectDataImportListener` await. Both triggers are symmetric pure functions with contract tests. Validated live: `Maven import finished: 136 project(s), 140 module(s)` on Keycloak.

test-integration/src/main/kotlin/com/jonnyzzz/mcpSteroid/integration/infra/mcp-steroid-import.kt
jonnyzzz: Done in #200 — settle = `awaitConfiguration` completed (primary; coroutine-tracked work is invisible to indicators) + smart mode + zero running indicators over a 10-round quiet window; a frozen signature (flags + task titles/text2/fractions) for 3 min while not configuring fails fast NAMING the stuck state; a script failure now genuinely fails the harness (the returned isError result was previously ignored). Validated on two live Keycloak imports.

test-integration/src/main/kotlin/com/jonnyzzz/mcpSteroid/integration/infra/mcp-steroid.kt
jonnyzzz: Done in #187 + #198 — the sniffing helper is gone; the request/parse layers throw the typed `McpRequestFailedError` (a `WaitAbortedError`/Error) directly, so the stop-the-waitFor behavior is implicit for every caller; `waitFor` itself only catches `Exception`.

test-integration/src/main/kotlin/com/jonnyzzz/mcpSteroid/integration/infra/mcp-steroid.kt
jonnyzzz: Done in #187 — the request layer throws automatically: `TransientMcpRequestException` (IDE busy, exit -1 → retried) vs `McpRequestFailedError` (terminal → breaks waitFor). `mcpExecuteCode` needs no try/catch; every terminal path reachable from a poll is typed (driven to closure by a 7-round adversarial quorum).

test-integration/src/test/kotlin/com/jonnyzzz/mcpSteroid/integration/infra/MavenImportSourceDownloadTest.kt
jonnyzzz: Done in #199 — Gradle enables source auto-download, the full available parity: IntelliJ's Gradle integration has no IDE-side javadoc auto-download (only `gradle.download.sources` is registered; `downloadJavadoc` is a build-script idea-plugin option), and *-sources.jar carries the docs. A regression test locks out the nonexistent `gradle.download.javadoc` id the first cut tried.

test-integration/build.gradle.kts
jonnyzzz: Done in a4b9c07f — aligned: `test.integration.ide.download.dir` now points at the ROOT-shared `build/ide-download` in both modules (it was per-module, so each IDE archive downloaded twice), the same convention as the dependency cache. Local warm archives (6.2 GB) migrated.

Issue #42, opened 2026-05-10
Default imports for the most-used PSI subpackages


andrii0lomakin:
## How this issue was triggered (real numbers from the log dataset)

In a 633-execution sample from the YTDB (YouTrackDB) project, scripts repeatedly import three PSI subpackages that aren't in the wrapper's default imports:

| Class                    | Sub-package              | Re-import count |
| ------------------------ | ------------------------ | --------------- |
| `GlobalSearchScope`      | `psi.search`             | 474             |
| `ReferencesSearch`       | `psi.search.searches`    | 223             |
| `ClassInheritorsSearch`  | `psi.search.searches`    | 63              |
| `PsiShortNamesCache`     | `psi.search`             | 28              |
| `MethodReferencesSearch` | `psi.search.searches`    | 26              |
| `PsiTreeUtil`            | `psi.util`               | 11              |

That's ~825 import lines for symbols in `psi.search.*`, `psi.search.searches.*`, `psi.util.*` — all *necessary* because those subpackages aren't in defaults. Adding three wildcards to the wrapper covers all of them.

Separately, ~1,100 imports in the dataset are *redundant* — agents re-import classes that *are* already covered by current defaults: `JavaPsiFacade` (460 redundant imports of `com.intellij.psi.JavaPsiFacade` when `com.intellij.psi.*` is in defaults), `PsiClass` (75 redundant), `Project` (47 redundant), `VfsUtil` (28 redundant), and so on. The skill guide says "default imports are provided" but doesn't enumerate them, so agents import defensively.

## Proposal

Add three wildcards to the wrapper's default imports:

```kotlin
import com.intellij.psi.search.*           // GlobalSearchScope, PsiShortNamesCache, …
import com.intellij.psi.search.searches.*  // ReferencesSearch, ClassInheritorsSearch, MethodReferencesSearch, OverridingMethodsSearch, …
import com.intellij.psi.util.*             // PsiTreeUtil, PsiUtil, CachedValuesManager, …
```

Three lines in the wrapper template — covers all ~825 currently-mandatory imports.

## Doc change

The skill guide (`mcp-steroid://prompt/skill`) currently says "Default imports are provided automatically. Add imports only when you need APIs outside the defaults." but doesn't enumerate what's covered. Add an explicit table of default imports — that's what kills the ~1,100 redundant imports today.

## Risk: name collisions

[...]

Issue #12, opened 2026-02-11
Make 'reason' parameter optional in steroid_execute_code


jonnyzzz:
## Problem

The `reason` parameter in `steroid_execute_code` is currently **required** (`ExecuteCodeToolHandler.kt:150`). This means agents must generate a reason string on every single tool call, consuming extra tokens for both:

1. **Input tokens**: The parameter description is long (~50 tokens) and instructs the agent to provide the full task description on the first call and a per-execution description on subsequent calls.
2. **Output tokens**: The agent must generate a reason string every time it calls execute_code, even for simple/obvious operations.

The `reason` parameter serves internal tracking purposes (stored in execution storage), but its value to the agent itself is zero — it's essentially metadata. Making it required forces token spend on every call.

## Current behavior

```kotlin
// ExecuteCodeToolHandler.kt:123-126
putJsonObject("reason") {
    put("type", "string")
    put("description", "IMPORTANT: On your FIRST call, provide the FULL TASK DESCRIPTION from the user...")
}

// Line 150 — reason is required
putJsonArray("required") {
    add("project_name")
    add("code")
    add("reason")    // <-- this should be optional
    add("task_id")
}
```

## Proposed Solution

1. Remove `"reason"` from the `required` array in the tool's JSON schema
2. Make the `reason` field nullable in `ExecCodeParams` (it's currently `val reason: String`)
3. Use a default like `""` or `null` when the agent doesn't provide it
4. Simplify the parameter description — remove the "IMPORTANT" instruction that asks for full task description on first call

This saves ~50 input tokens (shorter description) + whatever the agent generates as reason text per call.

## Related

Part of token optimization effort — see #11 (registry option to disable feedback tool).

jonnyzzz:
Kept open (deferred — not for 0.101). Confirmed still required (`CommonToolParams.reason()` is `.required()`, `ExecCodeParams.reason` non-null). Small additive change when picked up: drop `.required()` + default `reason = ""` (also on the wire DTO).

jonnyzzz:
Owner decision 2026-07-31: 'reason' stays REQUIRED on steroid_execute_code — the audit/telemetry value outweighs the per-call token cost. An implementation of the optional variant briefly rode PR #396 and was removed before merge per this decision. Closing as not planned.

PR #186, opened 2026-06-30
Fix exec_code compile-error line mapping drifting with the import count

matveidetonator:
CodeWrapperForCompilation wraps a submitted script with a preamble (default imports + boilerplate) before compiling, then maps wrapped-file line numbers back to the submitted script's lines so kotlinc diagnostics and stack-trace frames point at the line the agent actually wrote.

The mapping used hardcoded offsets (`15 + i` for imports, `23 + n + i` for code) computed for 12 default imports. The default-imports list has since grown (15, and 18 after #42 added the PSI subpackages), so every mapping was off by the difference: real diagnostics fell on unmapped wrapped lines and were shown to the agent verbatim (e.g. `input.kt:26` for a 2-line script), breaking "fix line N" self-correction loops.

Build the line mapping while emitting the wrapped code — record each user line's position as it is written — so the map can never drift from the generated layout again, regardless of how many default imports exist.

CodeWrapperForCompilationTest fed the stale wrapped line numbers (the same numbers the buggy map produced) and so never caught the drift; rewrite each case to derive the wrapped line from the actually-generated code and assert the remap round-trips. Closes item #12 in docs/agent-feature-review-2026-06.md.

### DIFF
--- a/kotlin-cli/src/main/kotlin/com/jonnyzzz/mcpSteroid/koltinc/CodeWrapperForCompilation.kt
+++ b/kotlin-cli/src/main/kotlin/com/jonnyzzz/mcpSteroid/koltinc/CodeWrapperForCompilation.kt
@@ -132,41 +132,64 @@ object CodeWrapperForCompilation {
         val importLines = extracted.importLines
         val otherLines = extracted.otherLines
 
-        // Build the wrapped code AND its line mapping in lock-step: record each user
-        // line's wrapped position as it is emitted. This is drift-proof — earlier the
-        // offsets were hardcoded (12 default imports → user code at line 23+N), so growing
-        // [defaultImports] silently shifted every user line and broke remapping (compiler
-        // errors then pointed at the generated line, not the submitted one). Never reintroduce
-        // magic offsets here; the only source of truth is the emission order below.
+        // Build the wrapped code AND its line mapping in lock-step: every line goes through
+        // [WrappedLineEmitter.emit], which appends it and advances the wrapped-line counter
+        // together. This is drift-proof — earlier the offsets were hardcoded (12 default
+        // imports → user code at line 23+N), so growing [defaultImports] silently shifted every
+        // user line and broke remapping (compiler errors then pointed at the generated line,
+        // not the submitted one). Because the counter can only move inside emit(), the mapping
+        // can never drift from the generated layout. Never reintroduce magic offsets here.
         val mapping = mutableMapOf<Int, Int>()
-        var wrappedLine = 0
         val wrappedCode = buildString {
-            defaultImports.forEach { appendLine(it); wrappedLine++ }
-            appendLine(); wrappedLine++
-            appendLine("//imports from the submitted code"); wrappedLine++
+            val out = WrappedLineEmitter(this)
+            defaultImports.forEach { out.emit(it) }
+            out.emit()
+            out.emit("//imports from the submitted code")
             importLines.forEachIndexed { i, line ->
-                appendLine(line); wrappedLine++
-                mapping[wrappedLine] = extracted.importLineNumbers[i]
+                out.emit(line)
+                mapping[out.line] = extracted.importLineNumbers[i]
             }
-            appendLine(); wrappedLine++
-            appendLine("class $clazzName {"); wrappedLine++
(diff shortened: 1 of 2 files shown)


jonnyzzz (approved):

kotlin-cli/src/main/kotlin/com/jonnyzzz/mcpSteroid/koltinc/CodeWrapperForCompilation.kt
```diff
@@ -132,59 +132,39 @@ object CodeWrapperForCompilation {
...
-            appendLine("  fun $methodName(builder : $scriptBuilderFqn) { ")
-            appendLine("    builder.$addBlockName { ${methodName}_code() }")
-            appendLine("  }")
-            appendLine("  suspend fun $scriptContextFqn.${methodName}_code() {")
-            appendLine("    //the rest of submitted code")
-            otherLines.forEach { append("    ").appendLine(it) }
-            appendLine("  }")
-            appendLine("}")
+            defaultImports.forEach { appendLine(it); wrappedLine++ }
+            appendLine(); wrappedLine++
```
jonnyzzz: I think you can create an val builder = object {
  fun appendLine(line: String) {
     parentBuilder.appendLine(line)
     wrappedLine++
     assertLineIsOnlyOneLine(line) //or do better
  }

Once done that way, we keep your wrapped lines, and actually build the mapping map implicitly and much more error-prone way.

WDYT?

kotlin-cli/src/test/kotlin/com/jonnyzzz/mcpSteroid/koltinc/CodeWrapperForCompilationTest.kt
```diff
@@ -16,13 +42,7 @@ class CodeWrapperForCompilationTest {
...
 
         val result = CodeWrapperForCompilation.wrap("Test", code)
-        val mapping = result.lineMapping
-
-        // "import foo.Bar" is on original line 1.
-        // In wrapped code, user imports start at line 15 (after 12 default imports + empty + comment).
-        // So wrapped line 15 should map to original line 1.
-        val remapped = mapping.remapCompilerOutput("input.kt:15:1: error: unresolved")
-        assertEquals("input.kt:1:1: error: unresolved", remapped)
+        assertRemaps(result, "import foo.Bar", originalLine = 1, col = 1)
```
jonnyzzz, line 55: This is a nice API, but I would be closer to the original lines, we need to check that code + line maps correctly with the different ammount of imports. 

The risk of the `assertRemaps` is that is can be so easy to tests slightly different scenarios of itself, and can slip. 

For the same of correctness it's always nice to add more tests to the original code and check if our fixes help fixing all cases, e.g. as TDD suggests

kotlin-cli/src/test/kotlin/com/jonnyzzz/mcpSteroid/koltinc/CodeWrapperForCompilationTest.kt
```diff
@@ -2,10 +2,36 @@
 package com.jonnyzzz.mcpSteroid.koltinc
 
 import org.junit.Assert.assertEquals
+import org.junit.Assert.assertTrue
 import org.junit.Test
 
 class CodeWrapperForCompilationTest {
 
+    /**
```
jonnyzzz, line 10: I would like to suggest a test for `inline fun` magic of Kotlin, when we call an include function it's budy is inlined and Kotlin uses the line numbers which are outside of the end of file. The main question is -- how we map such lines back, will it work correctly and still be understandable. This could be a followup task (so to avoid blocking you)

kotlin-cli/src/test/kotlin/com/jonnyzzz/mcpSteroid/koltinc/CodeWrapperForCompilationTest.kt
```diff
@@ -2,10 +2,36 @@
...
+     * Drift-proof remap assertion: locate where [snippet] physically lands in the generated
+     * code and verify that a compiler error on that wrapped line remaps back to [originalLine].
+     *
+     * Deriving the wrapped line from the emitted code — instead of hardcoding an offset like
```
jonnyzzz, line 14: What is the error is in the import line which we just added, how is that tested and reported?

PR #200, opened 2026-07-02
test-integration: settle imports by progress draining + stuck detection

jonnyzzz:
Review follow-up #4 from #185 ("instead check if progress is moving and it is not stuck"). **Stacked on #199** (`feat/unify-import-triggers`). 3 of 4 review-follow-up PRs.

## What
Settle phase extracted into pure `importSettleCode()` and reworked from a blind 10-s dumb-flag streak to progress semantics:
- **Settled** = `awaitConfiguration` completed within its bound (**primary** — coroutine-tracked work like the Maven source download is invisible to progress indicators; a timed-out await = "still configuring" = not quiet) + smart mode + zero running indicator tasks (secondary), over a 10-round quiet window.
- **Stuck** = observable signature (configuring/dumb flags + task titles/`text2`/fractions) frozen 3 min while not configuring → fails fast **naming the frozen state** (a blocked `awaitConfiguration` is tracked liveness, never stuck; hung tracked work falls to the 20-min deadline).
- The host now **checks the returned script result** — a script-level failure (trigger rethrow, STUCK) previously came back as an ignored isError result and was silently swallowed. Exec timeout 600 s → 30 min so it contains the budgets.

## Gates
- **Local:** unit contract tests green; script smoke-run on a live idle IDE (compiles, settles); **two full Keycloak Docker imports** (one per revision): `Maven import finished: 136 project(s), 140 module(s)` → `Settled — configuration drained + smart mode + no running background tasks` → red-code 1/25747 (the known false-positive). The only test failure is the unrelated missing local `ANTHROPIC_API_KEY` for the agent phase.
- **Quorum (adversarial, 3 rounds):** R1 both reviewers REJECT — swallowed failure channel, uncontainable timeout, `text2`-less signature, and the deepest: `getCurrentIndicators()` is blind to coroutine-tracked work, so `awaitConfiguration` completion must be the primary gate. All fixed. R3 codex **APPROVE**; claude R3 verdict to be posted as a PR comment when it lands.
- **Inspections:** changed files clean (0 meaningful).

### DIFF
--- a/test-integration/src/main/kotlin/com/jonnyzzz/mcpSteroid/integration/infra/mcp-steroid-import.kt
+++ b/test-integration/src/main/kotlin/com/jonnyzzz/mcpSteroid/integration/infra/mcp-steroid-import.kt
@@ -9,10 +9,11 @@ package com.jonnyzzz.mcpSteroid.integration.infra
  *   import-finished event (Gradle's `Observation.awaitConfiguration` can stay suspended after sync).
  * - NONE: nothing to trigger.
  *
- * Settle phase (shared): poll until the IDE stays out of dumb mode for a stable streak, re-running
- * `Observation.awaitConfiguration` each round. A single `awaitConfiguration` + `waitForSmartMode` snapshot
- * is NOT enough: import (and the always-on source/javadoc downloads) add library roots that re-trigger
- * indexing in waves, so a one-shot check can return in a gap between waves while indexing is still coming.
+ * Settle phase (shared, [importSettleCode]): wait for progress to DRAIN — smart mode plus zero running
+ * background progress indicators — and fail fast when progress FREEZES (stuck import). A one-shot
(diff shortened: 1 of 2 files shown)


jonnyzzz:
Quorum complete: the second reviewer also returned **VERDICT: APPROVE** on the final revision — both reviewers approve.

Non-blocking notes from that review, for the record:
- **Gradle residual risk:** if Gradle's `awaitConfiguration` stays suspended after sync, `configuring=true` every round → never quiet → the 20-min deadline burns and the run ends with an honest WARNING (vs the old code printing a false "Settled" under the same condition). Live evidence is Maven-only (Keycloak ×2) — worth watching one Gradle run of this revision.
- Indicator iteration order across snapshots isn't guaranteed stable; an order flip in a multi-indicator freeze resets the stuck clock (missed stuck → deadline catches it). Weakens fail-fast only; never a false failure.
- While `configuring=true`, rounds cost ~61 s, so the settling log prints ~once per 15 min in that mode — sparse but bounded.

Issue #177, opened 2026-06-29
runInspectionsDirectly hang in headless eval IDE: package-checker VulnerableCodeUsagesInspection blocks 10m (root cause: headless mode) — no-fix/tracking


jonnyzzz:
## Summary

In the 2026-06-26 steroid full eval run, a `steroid_execute_code` call invoking `runInspectionsDirectly()` **hung and pinned the IDE**. The agent had already finished its work (correct patch, Maven tests green), but the IDE never recovered: it sat blocked until the eval harness's hard wall-clock cap force-killed the whole run ~11 minutes later, *before* the scoring stage could run. The task therefore produced **no `metrics.json`** and was recorded as a failure/missing — even though the patch was complete and equivalent to the (passing) baseline.

I strongly suspect this is a **deadlock**, not merely "slow inspection work" or a missing timeout. Filing to track; root-cause investigation (thread dumps) is ongoing and will be added as comments. **No fix expected yet** — first we need to pin down where the deadlock comes from and how to guarantee it can't recur.

## Field evidence — task `dpaia__spring__petclinic__rest-24` (steroid full run)

The agent's last `steroid_execute_code` call ran inspections over 3 changed files:

```kotlin
val files = listOf(
    ".../service/UserServiceImpl.java",
    ".../security/BasicAuthenticationConfig.java",
    ".../security/PasswordEncoderConfig.java",
)
for (path in files) {
    val vf = findProjectFile(path)!!
    val problems = runInspectionsDirectly(vf)   // <-- never returns
    ...
}
```

The execution's `output.jsonl` stops at:

```
{"text":"[RUN] script"}
{"text":"Running inspections on UserServiceImpl.java..."}   <-- last line, never completes
```

That progress line is logged inside `runInspectionsDirectly` immediately before `waitForSmartMode()` and the `readAction { InspectionEngine.inspectEx(...) }` block (`McpScriptContextImpl.runInspectionsDirectly`). So the hang is in `waitForSmartMode()` or the inspection read action.

### IDE log (`logs/idea.log`) — the smoking guns

```
17:14:37 SEVERE #c.i.p.s.HeadlessSynchronizationService - timeout occurred while waiting for
         synchronization on librariesFqnVulnerabilitiesTimeout, timeout is 10m
         Caused by: kotlinx.coroutines.TimeoutCancellationException: Timed out waiting for 600000 ms
[...]

Issue #92, opened 2026-06-11
steroid_execute_code routes to the wrong project when two same-named projects are open (main checkout + git worktree)


jonnyzzz:
## Summary

When two projects with the same internal IntelliJ project name are open in one backend — a main checkout and its git worktree — `steroid_execute_code` routes BOTH suffixed `project_name` values to the **main checkout**. The worktree project is unreachable by name.

## Environment

- IntelliJ IDEA 2026.1.3 (IU-261.25134.95)
- MCP Steroid plugin 0.100-409f23a2, via devrig MCP
- macOS Darwin 25.5.0, 2026-06-11
- Reported by a Claude Code agent (dogfooding)

## Setup

Two projects open in the same backend:

| project_name | path |
|---|---|
| `marinator-feedback-XBqNNjRN` | `/Users/jonnyzzz/Work/marinator-feedback` (main checkout) |
| `marinator-feedback-MaAsLaQF` | `/Users/jonnyzzz/Work/marinator-feedback/.claude/worktrees/merge-point-detection` (git worktree) |

Both share the internal IntelliJ project name `marinator-feedback`.

## Observed

- `steroid_list_windows` shows the correct, distinct `projectPath` per window.
- `steroid_execute_code` with **either** suffixed `project_name` executes a script whose `project.basePath` == the **main checkout**. There is no `project_name` value that reaches the worktree project.
- Suffix stability is part of the problem: the worktree's `project_name` changed across re-registration — `merge-point-detection-MaAsLaQF` was briefly reported as "no longer present", and the project was later re-listed as `marinator-feedback-MaAsLaQF`.

## Consequence

PSI/inspection work silently ran in the **wrong project**: the worktree files resolved against the main checkout's project model — files landed in the root module with `inSourceContent=false`, producing mass false positives (`PackageDirectoryMismatch`, `UnusedSymbol` on essentially everything). The failure mode is silent — the script runs and returns plausible-looking results.

## Workaround used

Inside the script, ignore the routed `project` and look the real one up by path:

```kotlin
val project = ProjectManager.getInstance().openProjects
    .first { it.basePath?.contains(".claude/worktrees/merge-point-detection") == true }
```

…then drive `InspectionEngine` manually against that project.

## Expected

The suffixed `project_name` should uniquely route to the matching `Project` — disambiguate by project **path**, not by the (non-unique) internal project name. Suffixes should also be stable across re-registration.

## Related

[...]

PR #220, opened 2026-07-03
Lock-free project-name resolution — one hung read action no longer wedges every tool (#214)

### LINKED ISSUE #214: One hung read action wedges every steroid tool — project-name resolution takes a read lock it does not need
## Confirmed: the wedged-server behavior from #177 is read-lock starvation, and it hits every tool

In the 2026-06-26 eval run (`feature__service-45`), a `steroid_execute_code` running `runInspectionsDirectly` hung (the #177 headless package-checker deadlock) and the client timed it out at 16:50:08. Four seconds later the agent called `steroid_fetch_resource(mcp-steroid://prompt/test-skill)` — a static prompt fetch — and that also timed out at 60 s. The agent then abandoned the MCP entirely: zero steroid calls for the rest of the task (verified in the session jsonl).

The idea.log shows the full mechanism, step by step:

1. **16:49:20,896** — "Running 1430 inspections on LogRequestFilter.java". This execution never logs another line; the read action stays held until the harness kills the run at 17:05:21. It is uncancellable because `inspectEx` runs under a throwaway `EmptyProgressIndicator` (#179).
2. **16:49:23,125** — ~2 s later, a flood of `ProgressManagerImpl - checkCancelled is invoked while write-action is pending` starts (every ~100 ms). Something queued a write action behind the hung read action. The log does not name the writer — that part we cannot identify from logs alone.
3. **16:50:12,295** — the fetch_resource POST arrives and the handler *starts running* (`FetchResourceToolHandler - steroid_fetch_resource: ...` is printed). So the Ktor dispatcher, session manager, and tool registry were all fine — dispatch-level serialization and thread-pool exhaustion are ruled out.
4. The handler then blocks for exactly 60 s inside `buildPromptsContext` → `ProjectScopedToolHandler.resolveProject` → suspend `readAction {}`. At **16:51:12,291** the server logs `200 OK in 60002ms` with no response body line — the client disconnected at its own 60 s MCP timeout. There is no server-side timeout on this path.

Why the read action blocks is platform lock policy, not a plugin bug per se: `InternalReadAction.readLoop` parks every **new** suspend read action whenever a write action is pending (`community/platform/platform-impl/src/com/intellij/openapi/application/rw/InternalReadAction.kt:82-84`), and the parked reader resumes only after the write completes (`NestedLocksThreadingSupport.runWhenWriteActionIsCompleted`). The pending write can never start while the hung inspection holds its read permit. So: hung read action + one queued write = no new read actions, indefinitely.

`spring__petclinic__rest-24` shows the identical signature (inspections at 17:04:36, write-action-pending flood from 17:04:37,990); the agent there never issued another MCP call, so the wedge was not re-observed — but the mechanism is the same.

## The plugin-side problem: every tool takes a read action before doing anything

- `steroid_fetch_resource` serves **static** markdown. Its only IDE access is `resolveProject(projectName)` — and the result is **discarded** (`PromptsContextHandlerIJ.kt:10`); it is pure existence validation. That validation runs a suspend `readAction {}` (`ProjectScopedToolHandler.kt:28`).
- `steroid_list_projects` / `steroid_list_windows` take the same `readAction` in `describeSelfBackend` (`ListProjectsToolHandler.kt:47`).
- `execute_code`, `execute_feedback`, `take_screenshot`, `input` all route through `resolveProject`; `open_project` has its own `readAction` for the already-open check (`OpenProjectToolHandler.kt:37`).

So one stuck script turns the whole tool surface into 60 s dead timeouts, and the agent (rationally) gives up on the IDE.

None of these read actions are needed. `ProjectManager.getOpenProjects()` is explicitly documented as usable **outside** a read action if you check `Project.isDisposed()` per project (`community/platform/projectModel-api/src/com/intellij/openapi/project/ProjectManager.java:70-74`). `projectNameFor` only reads `project.name` and `project.basePath` — plain getters, no PSI/model access.

## Design

1. **Make project-name resolution lock-free.** In `ProjectScopedToolHandler.resolveProject`, `describeSelfBackend`, and `OpenProjectToolHandler`'s already-open check: drop `readAction {}`, call `ProjectManager.getInstance().openProjects` directly, and filter `!it.isDisposed`. After this, `fetch_resource`, `list_projects`, `list_windows`, `take_screenshot`, and `input` keep working through any read-lock starvation.
2. **Diagnose instead of dying silently (optional, defense in depth).** Where a tool-handler path still genuinely needs a read action, wrap it in a short `withTimeout` and return a `ToolCallErrorException` naming the in-flight execution ids ("a write action is queued behind a long-running read action; execution eid_... is still running") instead of letting the client time out blind.
3. **Relationship to #179.** The cancellable-indicator fix clears the wedge at its source: once the pending write (or the script timeout) can cancel the inspection read action, the lock releases within one `checkCanceled`. But this issue is still worth fixing independently — any *legitimately* long read action plus one queued write starves the trivial endpoints for its whole duration, even after #179.

Regression test: on a background thread, hold a read action on a latch; queue an asynchronous write action; then call `resolveProject` (and the fetch_resource handler) under a short timeout and assert they complete.

## What we could and could not verify from the logs

Verified: the full service-45 timeline (session jsonl + idea.log timestamps match to the millisecond); the write-action-pending flood during the hang in both wedged tasks; that the fetch_resource handler entered immediately and blocked in the read action, ruling out HTTP/dispatch causes; the platform park-new-readers-behind-pending-write behavior in the sources; that no server-side timeout exists on this path (the 60 s is the client's).

Not verifiable from logs: which component queued the write action (ProgressManagerImpl does not name the writer; dependency-model collection started at 16:49:22, but that is circumstantial); that `list_projects` specifically would have timed out (the agent never called it during the wedge — the claim is by code-path identity with fetch_resource); no thread dump exists from the wedge window.

## Acceptance

- With a read action held and a write action queued (simulated), `steroid_fetch_resource` and `steroid_list_projects` respond normally.
- `resolveProject` and `describeSelfBackend` contain no `readAction {}`.
- Existing not-found error behavior (listing available `project_name` keys) unchanged.
- Next eval run: a hung `execute_code` no longer produces follow-on timeouts on trivial tools (the service-45 signature — consecutive 60 s timeouts on unrelated tools — does not recur).

## Decision

Ship the lock-free resolution only (design item 1). The bounded-readAction diagnostic (item 2)
is deferred until after the cancellable-indicator fix lands and the next eval run shows whether
any wedge signature remains. Disposal race: project-scoped handlers re-check `isDisposed` at use.

jonnyzzz:
Fixes #214. Companion to #177/#179.

One hung read action + one queued write action starves every new read action (platform lock fairness) — and every steroid tool took a read action before doing anything, so a single stuck script turned the whole tool surface into 60s dead timeouts (the confirmed `service-45` wedge: hung `execute_code`, then even a static `fetch_resource` timed out).

The read actions were never needed:

- `ProjectScopedToolHandler.resolveProject`, `describeSelfBackend`, and `open_project`'s already-open check now read `ProjectManager.getInstance().openProjects` directly with `!isDisposed` filtering — the platform documents this as safe outside a read action (`ProjectManager.java:70-74`). `projectNameFor` reads only `name`/`basePath` getters.
- Not-found error message stays byte-identical; the #92 opaque-key matching is preserved; disposal races surface via `AlreadyDisposedException` on next use in project-scoped handlers (documented contract).
- The bounded-readAction diagnostic (issue design item 2) is deferred per the issue decision.

Regression test (`ProjectResolutionLockFreeTest`) reproduces the wedge for real: a raw thread holds a read action, a write action is queued on the EDT (verified via public `isWriteActionPending`), then full HTTP `steroid_fetch_resource` + `steroid_list_projects` calls are asserted to complete **while the write is still pending** — the exact `service-45` signature, now impossible.

Verification: full `:ij-plugin:test` green. 3-agent quorum review: codex APPROVE, claude APPROVE, gemini APPROVE — zero required changes. Transcripts in the private eval-logs repo.

### DIFF
--- a/ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/server/ProjectScopedToolHandler.kt
+++ b/ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/server/ProjectScopedToolHandler.kt
@@ -1,7 +1,6 @@
 /* Copyright 2025-2026 Eugene Petrenko (mcp@jonnyzzz.com); Copyright 2025-2026 JetBrains. Use of this source code is governed by the Apache 2.0 license. */
 package com.jonnyzzz.mcpSteroid.server
 
-import com.intellij.openapi.application.readAction
 import com.intellij.openapi.components.Service
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.project.ProjectManager
@@ -11,7 +10,7 @@ import com.jonnyzzz.mcpSteroid.mcp.ToolCallErrorException
  * Shared base for `*ToolHandlerIJ` implementations that resolve a
  * `projectName: String` argument to an open IDE [Project].
  *
- * Centralizes the read-action lookup and the not-found error message
+ * Centralizes the lock-free lookup and the not-found error message
  * (which lists the currently-open `project_name` routing keys to help the
  * caller correct a typo) so each handler only writes the project-scoped body.
  */
@@ -22,21 +21,29 @@ class ProjectScopedToolHandler {
      * [ToolCallErrorException] if no project matches — `McpToolRegistry.callTool`
      * catches it and turns it into an MCP error result naming the missing project
      * plus the list of currently-open `project_name` values.
+     *
+     * Deliberately lock-free (#214): a suspend `readAction {}` here parks behind any
+     * pending write action, so one hung read action (e.g. a stuck inspection script)
+     * would wedge EVERY steroid tool for the duration. [ProjectManager.getOpenProjects]
+     * is documented as safe outside a read action provided each project is checked for
+     * [Project.isDisposed] before use; [projectNameFor] only reads plain getters
+     * (`name`, `basePath`). The returned project can still be disposed concurrently —
+     * project-scoped handlers re-check [Project.isDisposed] at use.
      */
     @Throws(ToolCallErrorException::class)
-    suspend fun resolveProject(projectName: String): Project {
-        val (project, availableProjectNames) = readAction {
-            // (project_name, name, project) per open project — match by the OPAQUE project_name first,
-            // then by the raw folder name ONLY when unambiguous (exactly one match), never first-match,
-            // so two same-named projects (e.g. a checkout and its git worktree) can't silently collapse
-            // to the wrong one (#92).
-            val triples = ProjectManager.getInstance().openProjects.map { Triple(projectNameFor(it), it.name, it) }
-            val resolved = triples.firstOrNull { it.first == projectName }?.third
-                ?: triples.singleOrNull { it.second == projectName }?.third
-            // Surface the sorted project_name routing keys (not raw names) — that is the key callers must pass.
-            resolved to triples.map { it.first }.sorted()
-        }
-        return project ?: throw ToolCallErrorException(
+    fun resolveProject(projectName: String): Project {
+        // (project_name, name, project) per open project — match by the OPAQUE project_name first,
+        // then by the raw folder name ONLY when unambiguous (exactly one match), never first-match,
+        // so two same-named projects (e.g. a checkout and its git worktree) can't silently collapse
+        // to the wrong one (#92).
+        val triples = ProjectManager.getInstance().openProjects
+            .filter { !it.isDisposed }
+            .map { Triple(projectNameFor(it), it.name, it) }
+        val resolved = triples.firstOrNull { it.first == projectName }?.third
+            ?: triples.singleOrNull { it.second == projectName }?.third
+        // Surface the sorted project_name routing keys (not raw names) — that is the key callers must pass.
+        val availableProjectNames = triples.map { it.first }.sorted()
+        return resolved ?: throw ToolCallErrorException(
             "Project not found: \"$projectName\". Available project_name values: ${availableProjectNames.joinToString()}"
         )
     }
--- a/ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/server/ProjectScopedToolHandler.kt
+++ b/ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/server/ProjectScopedToolHandler.kt
@@ -10,7 +10,7 @@ import com.jonnyzzz.mcpSteroid.mcp.ToolCallErrorException
  * Shared base for `*ToolHandlerIJ` implementations that resolve a
  * `projectName: String` argument to an open IDE [Project].
  *
- * Centralizes the lock-free lookup and the not-found error message
+ * Centralizes the read-action lookup and the not-found error message
  * (which lists the currently-open `project_name` routing keys to help the
  * caller correct a typo) so each handler only writes the project-scoped body.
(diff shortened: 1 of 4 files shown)


ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/server/ProjectScopedToolHandler.kt
```diff
@@ -22,21 +21,29 @@ class ProjectScopedToolHandler {
...
-            // to the wrong one (#92).
-            val triples = ProjectManager.getInstance().openProjects.map { Triple(projectNameFor(it), it.name, it) }
-            val resolved = triples.firstOrNull { it.first == projectName }?.third
-                ?: triples.singleOrNull { it.second == projectName }?.third
-            // Surface the sorted project_name routing keys (not raw names) — that is the key callers must pass.
-            resolved to triples.map { it.first }.sorted()
-        }
-        return project ?: throw ToolCallErrorException(
+    fun resolveProject(projectName: String): Project {
+        // (project_name, name, project) per open project — match by the OPAQUE project_name first,
```
jonnyzzz: do a much smaller change here -- just replace `runAction` to `run` function, in such a way there will be no other changes.

jonnyzzz:
Done — minimal diff as requested: the three sites keep their exact signatures, structure, matching logic, and error messages from main; the only change is `readAction {` → `run {` (plus the now-unused import removal and a one-line rationale comment). The regression test is unchanged in substance and still proves the wedge fix: with a read action held and a write action pending, `resolveProject`, `describeSelfBackend`, and full HTTP `fetch_resource`/`list_projects` calls complete while the write is still pending.

One nuance for the record: `ProjectManager.getOpenProjects()`'s javadoc asks callers outside a read action to check `Project.isDisposed()` per project. With the wrapper-swap-only change that check is not added; a project disposed mid-resolution surfaces as `AlreadyDisposedException` on first use in the handler (same failure the read-action version had for a project disposed right after the action). Happy to add the one-line `.filter { !it.isDisposed }` if you want the contract followed literally — it would be the only non-wrapper change.

**readAction usage double-check** (full sweep of `ij-plugin/src/main`):
- `server/` tool-handler paths: **zero** read actions remain — every trivial endpoint is now starvation-proof.
- All remaining sites are `execution/` script-context helpers guarding genuine model access: daemon highlights + `getDocument` (`McpScriptContextImpl:363,422,427`), `runInspectionsDirectly` PSI+inspectEx (`:466` — cancellable since #229), the script-API `readAction/writeAction/smartReadAction` wrappers themselves (`:531`), `findPsiFile`/`findProjectPsiFile` PSI resolution (`:570,:624`), `findProjectFiles` VFS iteration (`:602`). These are correct: they run inside user scripts where write-preemption fairness is intended, and a slow one can only stall its own execution — never the tool surface.
- `vision/`, `storage/`, `mcp-http`: no read actions.

Branch rebased onto main (post-#219). Targeted suites green (`ProjectResolutionLockFreeTest`, `FetchResourceToolTest`, `McpServerIntegrationTest`).

Issue #179, opened 2026-06-29
runInspectionsDirectly passes a throwaway EmptyProgressIndicator() to InspectionEngine.inspectEx — should use the execution's cancellable progress indicator


jonnyzzz:
## Summary

`runInspectionsDirectly` calls `InspectionEngine.inspectEx(...)` with a **freshly constructed `EmptyProgressIndicator()`** instead of the cancellable progress indicator the execution already runs under. This severs cancellation: `ProgressManager.checkCanceled()` inside inspections — and any `runBlockingCancellable {}` an inspection performs — derives its cancellation from the *current* progress indicator, which is now an empty one that is **never cancelled**. As a result, a `steroid_execute_code` timeout/cancellation cannot interrupt the inspection sweep, and a single misbehaving (slow / blocking / hanging) inspection pins the read-action worker indefinitely.

This is the in-our-control half of the headless-mode deadlock in #177 (where `VulnerableCodeUsagesInspection` blocks for 10 minutes and the script timeout can't cancel it). But it's a general cancellation-correctness bug, independent of that specific inspection.

## Location

`ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/execution/McpScriptContextImpl.kt:511` — the only `EmptyProgressIndicator()` in the plugin:

```kotlin
return readAction {                                    // suspend readAction — coroutine/job-bound
    ...
    val results = InspectionEngine.inspectEx(
        toolWrappers, psiFile, psiFile.textRange, psiFile.textRange,
        false, false, true,
        EmptyProgressIndicator(),                      // <-- discards the execution's cancellation context
        PairProcessor.alwaysTrue()
    )
    ...
}
```

## Why this is wrong

We already execute under a cancellable context:

- `ScriptExecutor.executeCodeBlocks` wraps the script body in `withTimeout(exec.timeout.seconds)` (`ScriptExecutor.kt:237`) — a cancellable coroutine/Job.
- `runInspectionsDirectly` runs `inspectEx` **inside `readAction {}`** (`com.intellij.openapi.application.readAction`, the suspend, job-bound read action), which already installs a job-bound progress indicator.

`inspectEx` then *overrides* that ambient indicator with a throwaway `EmptyProgressIndicator()`, so:

- `checkCanceled()` calls inside inspections never observe our timeout;
- a nested `runBlockingCancellable {}` (e.g. the package-checker vulnerability sync in #177) can only be ended by its own internal timer, not ours;
[...]

Issue #221, opened 2026-07-03
Compiler-error line remapping is off by 4: LineMapping hardcodes 12 default imports, the list has 16


jonnyzzz:
Found during #213 implementation review.

`CodeWrapperForCompilation`'s `LineMapping` uses hardcoded offsets (user imports at line 15, code at 23+N) that assume **12** default imports — but the list has had 15 since `cc6102aa` and now has 16 (#213 added `ProgressIndicator`). Runtime compiler-error line remapping is off by 4 lines: the `input.kt:N` coordinates agents see in compile errors point 4 lines away from their actual code.

`CodeWrapperForCompilationTest` stays green because it pins the map's own constants, not the real wrapped layout — the test verifies the bug.

Fix: derive the offsets from `defaultImports.size` (user imports = size+3, code = size+11+N) and update the ~20 test expectations to be layout-derived too, so the next import addition cannot silently skew the mapping again.

Why it matters: the #156 work made error line references (`input.kt:N`) a primary diagnostic for agents; a 4-line skew sends them to the wrong statement.

jonnyzzz:
@matveidetonator is that fixed in `main` already?

jonnyzzz:
Closing after verification against current main. PR #186 is merged (c3dca9b8; implementation commits 503de4b6 and 45afd6fb): line mapping is derived through the emitted wrapper layout and guarded by drift-resistant tests. Open PR #231 is an obsolete earlier implementation and is not needed for this issue to be complete.

jonnyzzz:
Fixed in **v0.102**. Release notes: https://devrig.dev/releases/0.102/ · GitHub release: https://github.com/jonnyzzz/mcp-steroid/releases/tag/v0.102

PR #229, opened 2026-07-03
Run steroid_execute_code as a visible, cancellable IDE background task (#213)

### LINKED ISSUE #213: Give scripts a cancellable progress indicator on McpScriptContext; use it in runInspectionsDirectly (#179)
## Problem

`runInspectionsDirectly` hands `InspectionEngine.inspectEx` a fresh `EmptyProgressIndicator()` (`McpScriptContextImpl.kt:513`). `inspectEx` installs that indicator on its worker threads (`JobLauncher.invokeConcurrentlyUnderProgress`), so every `checkCanceled()` inside the inspections consults an indicator that nobody ever cancels. The script timeout (`withTimeout(exec.timeout.seconds)`, `ScriptExecutor.kt:237`) cancels the coroutine job tree — including the child read-job that suspend `readAction` installs — but a job cancellation never reaches that indicator. The sweep runs to the end no matter what.

Consequences, measured:

- #177 / the 2026-06-26 eval: `rest-24` hung in `runInspectionsDirectly` for ~11 minutes and the run was killed before scoring, with a complete correct patch lost.
- The 2026-06-29 full run confirmed at scale: `rest-24` and `service-45` wedged out of scoring entirely; in `service-45` the hung `execute_code` also wedged the server — a trivial `fetch_resource` sent 4 s later timed out too (REVIEW.md, "Timeout class").
- The reproducer on `repro/rest-24-inspection-hang`: a 2 s script timeout is *reported* but `executeWithProgress` does not return for ~12 s, because `withTimeout` cannot resume until the un-cancellable inspection finishes.

The fix suggested in #179 (`ProgressManager.getGlobalProgressIndicator() ?: EmptyProgressIndicator()`) does not work: suspend `readAction` installs a job into the thread context, not an indicator (`platform-impl/.../rw/cancellableReadAction.kt:26-30`), so the global indicator is null there and the fallback recreates the bug.

## Design

**1. New context member: `val progressIndicator: ProgressIndicator` on `McpScriptContext`.**

A per-execution, cancellable indicator. Scripts pass it to any IntelliJ API that takes an explicit `ProgressIndicator`. It is cancelled the moment the execution's coroutine job is cancelled — script timeout, client cancel, or disposal.

Implementation: a small plugin-side class extending `EmptyProgressIndicatorBase` + `StandardProgressIndicator` with a sticky volatile cancel flag. Not `EmptyProgressIndicator` itself: its `start()` *clears* the cancellation flag (`EmptyProgressIndicator.java:34-37`), so any consumer that calls `ProgressManager.runProcess(indicator)` would silently un-cancel it. The platform hit the same trap and solved it the same way with its internal `BridgeJobIndicatorBase` — we replicate those ~15 lines because that class is `@ApiStatus.Internal`. `EmptyProgressIndicatorBase.checkCanceled()` is final and consults our overridden `isCanceled()`, so the sticky flag is honored everywhere. Constructor is `@ApiStatus.Obsolete` — acceptable: there is no non-obsolete public way to construct an indicator, and bridging to indicator-driven blocking APIs is exactly the use the platform docs sanction.

**2. Wire it to the execution job in `ScriptExecutor.executeCodeBlocks`.**

`Job.invokeOnCompletion(onCancelling = true)` is `@InternalCoroutinesApi`, so use a plain watcher coroutine inside the `withTimeout` scope:

```kotlin
withTimeout(exec.timeout.seconds) {
    val watcher = launch(start = CoroutineStart.UNDISPATCHED) {
        try { awaitCancellation() } finally { context.progressIndicator.cancel() }
    }
    try { /* run blocks */ } finally { watcher.cancel() }
}
```

When the timeout (or an outer cancel) fires, the watcher's suspended continuation is resumed immediately — it does not wait for siblings — and cancels the indicator; the blocking sweep observes it within one `checkCanceled()` and unwinds, letting `withTimeout` actually return. Normal completion also cancels the indicator via the watcher's `finally`; the indicator's lifetime is the script body, nothing consumes it afterwards (an implementation may keep a flag to skip that if it reads cleaner).

**3. `runInspectionsDirectly` passes `progressIndicator` instead of `EmptyProgressIndicator()`** (and the import goes away). Suspend `readAction` is kept — it stays write-action-preemptible; the indicator only adds the missing cancellation path into the blocking sweep. This closes #179 and turns the `rest-24`/`service-45` wedge into a prompt, clean timeout error.

**4. KDoc on `McpScriptContext.progressIndicator`:** the script executes under this indicator; it is cancelled when the execution is cancelled (timeout included); pass it to IntelliJ APIs with a `ProgressIndicator` parameter; never construct `EmptyProgressIndicator()` in a script. Also note the ambient story honestly: indicators are thread-bound, so a suspend body cannot have one installed globally — implicit `ProgressManager.checkCanceled()` in blocking sections already observes the job (suspend `readAction` installs a job-bound thread context), and for blocking code that needs an *ambient* indicator the public bridge is `coroutineToIndicator { indicator -> ... }`. Do not mention `jobToIndicator`/`runUnderIndicator`: the first is `@ApiStatus.Internal`, the second no longer exists in the platform.

**5. Prompt corpus.** Replace the taught `EmptyProgressIndicator()` construction with the context's `progressIndicator`:

- `prompts/src/main/prompts/ide/inspect-and-fix.md` — both fences (lines 21/69 and 149/184), drop the import.
- `prompts/src/main/prompts/ide/find-duplicates.md` — the clone-scan fence (lines 166/222), drop the import.
- `prompts/src/main/prompts/skill/coding-with-intellij-context-api.md` — add `progressIndicator` to "Core Properties" and to the "Real helpers vs invented names" list, plus one line in the `runInspectionsDirectly` section.

No stub work needed for compilation tests: `KtBlockCompilationTestBase` inlines the real `McpScriptContext.kt` source (`KtBlockCompilationTestBase.kt:269`), and `CodeWrapperForCompilation` wraps user code as an extension on that interface, so the new property resolves in every fence; `com.intellij.openapi.progress.ProgressIndicator` is core-api and present in every IDE distribution.

**Gate note (PHILOSOPHY.md Tenet 3):** this is a new `McpScriptContext` member, which is gated. Justification: the execution's cancellation is plugin-internal state — no IntelliJ API call and no recipe can hand a script an indicator tied to *this execution's* timeout. The reviewer quorum should still run before implementation.

## Acceptance

- A test (adapt `RunInspectionsDirectlyTimeoutTest` from `repro/rest-24-inspection-hang`): register an inspection that spins with `checkCanceled()` for ~10 s, run `runInspectionsDirectly` in a script with `timeout = 2`; the execution reports the timeout error **and** `executeWithProgress` returns in roughly 2 s, not ~12 s.
- A test that `progressIndicator.isCanceled` is false during a normal run and the property is usable from script code (compiles through `CodeButcher` wrapping).
- `EmptyProgressIndicator` no longer appears in `ij-plugin/src/main` nor in any `prompts/src/main/prompts` kotlin fence.
- `./gradlew :ij-plugin:test` green; `:prompts:test --tests '*KtBlock*'` green (kotlin fences change, so the matrix run is required); `*MarkdownArticleContract*` green.
- Platform-API check: no new `@ApiStatus.Internal`/`@Experimental` usages (verified below; `EmptyProgressIndicatorBase`'s `@Obsolete` constructor is the single, documented exception).

## API verification (IntelliJ community sources, master @ 6879a98e7d2e2)

| API | Status | Evidence |
|---|---|---|
| `coroutineToIndicator((ProgressIndicator) -> T)` | public, stable | `core-api/.../progress/coroutines.kt:497` (no annotation; 0-arg overload at :467 is `@Deprecated(HIDDEN)`) |
| `jobToIndicator` | `@ApiStatus.Internal` — do not use | `coroutines.kt:572-573` |
| `blockingContextToIndicator` | `@ApiStatus.Internal` — do not use | `coroutines.kt:527-530` |
| `runUnderIndicator` | removed from the platform | grep over community returns nothing |
| `EmptyProgressIndicatorBase` | public abstract; ctor `@Obsolete`; `start()` keeps cancellation | `EmptyProgressIndicatorBase.java:38-53,73-83` |
| `EmptyProgressIndicator` | ctor `@Obsolete`; `start()` **clears** cancellation — avoid | `EmptyProgressIndicator.java:24-37` |
| `StandardProgressIndicator` | public | `StandardProgressIndicator.java:13` |
| `InspectionEngine.inspectEx` (9-arg, `PairProcessor`) | public (the overload already in use) | `InspectionEngine.java:116`; 10-arg overload :129-131 is `@Internal`, 5-arg :106 deprecated-for-removal |
| suspend `readAction` | public; installs a job, not an indicator | `core-api/.../application/coroutines.kt:32`; `platform-impl/.../rw/cancellableReadAction.kt:26-30` |
| kotlinx `awaitCancellation`, `CoroutineStart.UNDISPATCHED` | public kotlinx | (`invokeOnCompletion(onCancelling=true)` is `@InternalCoroutinesApi` — that is why the watcher pattern is used) |

## Related

Closes #179. Mitigates the in-our-control half of #177 (headless root cause tracked there). Eval evidence: `mcp-steroid-logs/2026-06-29/REVIEW.md` (rest-24 + service-45 wedged pre-scoring; wedged-server timeline in service-45).

## Decisions

- Scope: indicator only — no bounded sub-timeout for runInspectionsDirectly in this issue.
- Name: `progressIndicator`.
- PHILOSOPHY Tenet-3 gate (new context member): satisfied by the 3-agent quorum review this
  PR goes through before merge.
- Implements and closes #179.

jonnyzzz:
Fixes #213. Mitigates #179 at the execution boundary (see note).

## What

Every `steroid_execute_code` execution now runs as a **visible IDE background task** — "devrig -- MCP Steroid task" in the status bar — via the platform's coroutine-native progress API:

```kotlin
withBackgroundProgress(project, "devrig -- MCP Steroid task", cancellable = true) {
    // the existing execution, unchanged — withTimeout inside
}
```

One wrapper call in `ExecutionManager`. No custom indicator, no watcher coroutines, no manual cancellation wiring — the platform already handles the full loop (verified in `PlatformTaskSupport`): the task's UI Cancel cancels the coroutine context, and the task is removed from the status bar in a `finally` whenever the coroutine ends — completion, failure, script timeout, or **client disconnect**: `executeWithProgress` is a plain suspend fun with no scope hop, so the execution is a child of the Ktor request coroutine and structured concurrency unwinds it. Pinned by `testParentCoroutineCancellationUnwindsExecution` (cancels the parent mid-script, asserts prompt unwind).

Also included: **`ProcessCanceledException` rethrow** in `ScriptExecutor`'s catch chain (control-flow exception; defensive — PCE extends CE on current platforms).

The wrapper is purely additive: it nests around the existing `coroutineScope`, which keeps owning the sub-task structure (the response builder's supervised storage writes, progress streams) exactly as before — no other line of the execution path changes. The #221 line-mapping fix moved to its own PR.

## #179 note

`runInspectionsDirectly` is deliberately left unchanged in this PR. The in-read-action indicator bridge the platform uses for `inspectEx` (`jobToIndicator`) is `@ApiStatus.Internal`; the public-only alternatives all change the read-action shape. With this PR a stuck sweep is bounded by the visible task + child-of-request cancellation at the execution boundary; interrupting the sweep mid-flight stays open on #179.

## Verification

Full `:ij-plugin:test` green — every existing execution test now flows through the background-progress wrapper in the unit-test environment, plus the new disconnect-simulation test.

### DIFF
--- a/ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/execution/ExecutionManager.kt
+++ b/ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/execution/ExecutionManager.kt
@@ -6,6 +6,7 @@ import com.intellij.openapi.components.Service
 import com.intellij.openapi.components.service
 import com.intellij.openapi.diagnostic.thisLogger
 import com.intellij.openapi.project.Project
+import com.intellij.platform.ide.progress.withBackgroundProgress
 import com.jonnyzzz.mcpSteroid.mcp.ContentItem
 import com.jonnyzzz.mcpSteroid.mcp.ToolCallResult
 import com.jonnyzzz.mcpSteroid.mcp.builder
@@ -53,7 +54,16 @@ class ExecutionManager(
         exec: ExecCodeParams,
         mcpProgressReporter: McpProgressReporter,
     ): ToolCallResult {
-        return coroutineScope {
+        // The execution runs as a VISIBLE, cancellable IDE background task
+        // (com.intellij.platform.ide.progress.withBackgroundProgress — the coroutine-native
+        // progress API). Cancellation is fully platform-handled: the Cancel button cancels
+        // this coroutine (PlatformTaskSupport subscribes to the task entity and calls
+        // context.cancel()), and when the coroutine ends for ANY reason — completion,
+        // failure, script timeout, or the HTTP request coroutine being cancelled on client
+        // disconnect (this suspend fun is a child of the Ktor request coroutine; structured
+        // concurrency propagates) — the task disappears from the status bar automatically.
+        // No manual wiring is needed or wanted.
+        return withBackgroundProgress(project, "devrig -- MCP Steroid task", cancellable = true) { coroutineScope {
             val executionId = project.executionStorage.writeNewExecution(exec)
             withContext(CoroutineName("mcp-steroid-$executionId")) {
                 log.info("Starting execution $executionId-${exec.taskId}-${exec.reason}...")
@@ -121,7 +131,7 @@ class ExecutionManager(
 
                 builder.build()
             }
-        }
+        } }
     }
 
     private fun responseBuilder(parentScope: CoroutineScope, executionId: ExecutionId, mcpProgress: McpProgressReporter) = object : ExecutionResultBuilder {
--- a/ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/execution/ScriptExecutor.kt
+++ b/ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/execution/ScriptExecutor.kt
@@ -1,6 +1,7 @@
 /* Copyright 2025-2026 Eugene Petrenko (mcp@jonnyzzz.com); Copyright 2025-2026 JetBrains. Use of this source code is governed by the Apache 2.0 license. */
 package com.jonnyzzz.mcpSteroid.execution
 
+import com.intellij.openapi.progress.ProcessCanceledException
 import com.intellij.openapi.Disposable
 import com.intellij.openapi.application.EDT
 import com.intellij.openapi.application.ModalityState
@@ -252,6 +253,12 @@ class ScriptExecutor(
             resultBuilder.logRemappedException("Execution timed out", e, evalResult.lineMapping)
             resultBuilder.reportFailed("Execution timed out after ${exec.timeout} seconds")
         } catch (e: CancellationException) {
+            throw e
+        } catch (e: ProcessCanceledException) {
+            // Control-flow exception (Logger contract): rethrow, never report as a script
+            // failure. On current platforms PCE extends CancellationException, so the CE
+            // branch above already covers it — this branch is explicit defense for any
+            // PCE that does not.
             throw e
         } catch (t: Throwable) {
             // #156: never report an empty error. A messageless throwable (e.g. the bare
--- a/ij-plugin/src/test/kotlin/com/jonnyzzz/mcpSteroid/execution/ExecutionManagerTest.kt
+++ b/ij-plugin/src/test/kotlin/com/jonnyzzz/mcpSteroid/execution/ExecutionManagerTest.kt
@@ -10,6 +10,10 @@ import com.jonnyzzz.mcpSteroid.server.ModalMode
 import com.jonnyzzz.mcpSteroid.server.NoOpProgressReporter
 import com.jonnyzzz.mcpSteroid.setSystemPropertyForTest
 import com.jonnyzzz.mcpSteroid.testExecParams
+import kotlinx.coroutines.CompletableDeferred
+import kotlinx.coroutines.cancelAndJoin
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
 import kotlin.time.Duration.Companion.seconds
 
 /**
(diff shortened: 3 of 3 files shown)


jonnyzzz:

PR #230, opened 2026-07-03
Thread + coroutine dumps on execute_code timeout; result carries the folder path only (#215)

### LINKED ISSUE #215: On execute_code timeout, save a thread dump and a coroutine dump into the execution folder
## Why

When a `steroid_execute_code` run hits its timeout, we suspect a deadlock (stuck read action, EDT pin, background sync) — but we have nothing to diagnose it with after the fact. Issue #177 says it directly: "Thread dumps are needed to confirm what holds/awaits what." In the 2026-06-26 eval run, two tasks (`rest-24`, `service-45`) wedged the IDE after a steroid-call timeout until the harness killed the run before scoring; the hung window left no dumps behind.

We already capture thread dumps in the modal-dialog paths (`McpScriptContextImpl.captureThreadDump`, used at `syncDocuments-timeout`, `modal-monitor`, etc., and `ScriptExecutor.requireNonModalOrFail` writes `thread-dump-modality-gate.txt`). The one place we do NOT capture anything is the main script timeout — the exact suspected-deadlock case. And we never capture a coroutine dump anywhere, even though a stuck `runBlockingCancellable` or suspended job tree is only visible there.

Timing works in our favor: the #177 reproducer showed that after `withTimeout` fires, the stuck block keeps running (the read action is not cancellable from outside), so a dump taken inside the `TimeoutCancellationException` handler still shows the stuck frames.

## What

When `executeCodeBlocks` catches `TimeoutCancellationException` (`ScriptExecutor.kt:249`), capture:

- a full JVM thread dump — `com.intellij.diagnostic.ThreadDumper.dumpThreadsToString()` (public, already used by the plugin), and
- a kotlin coroutine dump — `com.intellij.diagnostic.dumpCoroutines(stripDump = false)` (top-level fun in `coroutineDumper.kt`, no `@ApiStatus` annotation; the platform installs the kotlinx debug probes at IDE startup by default, gated on `idea.enable.coroutine.dump` which defaults to `true`),

and write both into the execution's storage folder via `project.executionStorage.writeCodeExecutionData(...)` — that resolves to `.idea/mcp-steroid/eid_*/`, same as the existing `thread-dump-*.txt` files:

- `thread-dump-timeout.txt`
- `coroutine-dump-timeout.txt`

`dumpCoroutines()` returns `null` when the probes are not installed (property off, or a test application). In that case write a one-line note into `coroutine-dump-timeout.txt` ("coroutine dump unavailable: kotlinx debug probes not installed") so the folder is self-explaining. Do NOT call `DebugProbes.install()` ourselves — it is `@ExperimentalCoroutinesApi` in kotlinx and the platform owns that lifecycle.

The tool result must mention only the path, never the dump content (dumps are large). Extend the existing failure line to something like:

```
FAILED: Execution timed out after 60 seconds. Diagnostic thread + coroutine dumps stored at /path/to/project/.idea/mcp-steroid/eid_20260626T170431-task/
```

## Design

- Capture the two dump strings first with plain non-suspending calls (`ThreadDumper.dumpThreadsToString()`, `dumpCoroutines(stripDump = false)`), then write the files under `withContext(NonCancellable + Dispatchers.IO)`. Reason: the timeout case can race with an external cancellation of the whole tool call (client disconnect — see the `service-45` timeline where a follow-up `fetch_resource` also timed out), and `writeCodeExecutionData` internally does `withContext(Dispatchers.IO)`, which throws on entry if the caller's job is already cancelled. `NonCancellable` makes the write safe either way.
- Get the folder path for the message from `executionStorage.resolveExecutionDir(executionId)` (absolute — it resolves off `project.basePath`).
- Wrap the whole capture in `try/catch` that rethrows `CancellationException` and logs anything else via `log.warn`, mirroring the existing `captureThreadDump` — a failed dump must never mask the timeout report.
- Small refactor: extract the "write thread dump (+ coroutine dump) for reason X" logic into a shared helper so `McpScriptContextImpl.captureThreadDump` and `requireNonModalOrFail` can delegate to it. Adopting the coroutine dump in the modal-diagnostic sites is then a one-line change; keep it optional to limit scope.
- Note: `ThreadDumper.getThreadDumpInfo(...)`, which would produce a single combined dump, is `@ApiStatus.Internal` (ThreadDumper.java:61) — off-limits. Two separate public calls give the same information.

## Acceptance

- A `steroid_execute_code` run that times out leaves both `thread-dump-timeout.txt` and `coroutine-dump-timeout.txt` in its `.idea/mcp-steroid/eid_*/` folder.
- The MCP tool result contains the existing `FAILED: Execution timed out after N seconds` line extended with the absolute dump folder path, and none of the dump content (no thread stack frames, no coroutine dump header).
- When debug probes are not installed, the coroutine-dump file contains the one-line unavailability note; the execution result is otherwise unchanged.
- Dump capture survives a concurrent external cancellation of the tool call (NonCancellable write path).
- An `ij-plugin` test drives a script past a short timeout (same pattern as the #177 reproducer: a block doing non-interruptible sleep, `timeout = 1..2s`) and asserts both files exist and the result text mentions the path only.

## Related

- #177 — hung `runInspectionsDirectly`, run killed before scoring; explicitly asks for thread dumps from the hung window.
- #179 — the cancellation fix for the inspection path; this issue is the complementary diagnostics so the next hang is debuggable regardless of where it comes from.

## Cancellation flow (decided; platform-source verified)

Dumps fire on **timeout only** — but the timeout handler must also complete the cancellation
properly, so the platform can unwind the hang instead of leaving it running behind us:

1. On timeout, **cancel the execution's progress indicator** (from the McpScriptContext
   indicator work) — the platform's own bridge does exactly this:
   `job.invokeOnCompletion(onCancelling = true) { if (it is CancellationException) indicator.cancel() }`
   (`community/platform/core-api/src/com/intellij/openapi/progress/coroutines.kt:580-583`).
   With both the coroutine and the indicator cancelled, blocking platform code throws
   `ProcessCanceledException` at its next `checkCanceled()`, and suspend boundaries bridge
   PCE↔CancellationException automatically (`CeProcessCanceledException`, `coroutines.kt:173-174`).
2. **Handle PCE and the coroutine cancellation exceptions correctly** in ScriptExecutor's catch
   chain: PCE (including `CeProcessCanceledException`, which *extends* PCE) must be rethrown —
   never logged as an error, never reported as "Unexpected error". Today's `catch (t: Throwable)`
   can swallow a bare PCE; add the explicit rethrow.
3. Keep the change **minimalistic**: cancel indicator + capture dumps (NonCancellable) + report
   path + PCE rethrow. No new timeout machinery — `withTimeout` remains the single source of
   cancellation.

jonnyzzz:
Fixes #215. **Stacked on #229** (`fix/213-progress-indicator`) — merge that first; this PR's own delta is the last two commits.

## What

When a `steroid_execute_code` execution hits its timeout (the suspected-deadlock case), the plugin now:

1. **Cancels the execution's progress indicator first** (belt-and-braces over #229's watcher — with both the coroutine and the indicator cancelled, blocking platform code unwinds via `checkCanceled`/PCE and the platform's own CE bridging, per the flow decided in #215).
2. Captures a **full JVM thread dump** (`ThreadDumper.dumpThreadsToString`) and a **kotlin coroutine dump** (`dumpCoroutines` — platform-installed probes; a one-line unavailability note when absent) as strings, then writes `thread-dump-timeout.txt` + `coroutine-dump-timeout.txt` into the execution's storage folder under `NonCancellable + Dispatchers.IO` (required: the job is already cancelled).
3. Reports the failure with **only the absolute folder path appended** — never dump content — and only when something was actually written.

Dump capture can never mask the timeout report: each file writes independently, write failures preserve the captured dump in idea.log at WARN, CE is rethrown, everything else is swallowed at WARN.

**Shared helper adoption**: all five modal-path `captureThreadDump` sites and the modality gate now delegate to `captureDiagnosticDumps` — same file naming, plus coroutine dumps everywhere.

## Tests

- Timeout on a stuck non-interruptible thread (the #177 shape): both files exist, thread dump names the stuck thread, message carries the absolute dir and **no dump content**.
- Real failure injection (directories squat **both** dump filenames — no test seam): timeout still reported, and the message makes **no storage claim**.
- Shared-helper contract test for the modal-path sites.

## Verification

Full `:ij-plugin:test` green on the stack. 3-agent quorum: codex APPROVE, gemini APPROVE, claude APPROVE-WITH-CHANGES — all three required changes applied and re-tested (write isolation, honest no-claim message, KDoc). One human-note from implementation: unit-test JVMs don't install coroutine probes, so real coroutine-dump content should get a Docker/integration spot-check before the next eval campaign. Transcripts in the private eval-logs repo.


jonnyzzz:
Superseded by a clean rebuild on current main: the branch was stacked on the pre-rework #213 and carried removed APIs (McpExecutionProgressIndicator, the progressIndicator context member) plus committed review files. Replacement PR targets main with only the #215 dump-capture payload.

PR #231, opened 2026-07-03
Derive compile-error line-mapping offsets from defaultImports.size (#221)

### LINKED ISSUE #221: Compiler-error line remapping is off by 4: LineMapping hardcodes 12 default imports, the list has 16

jonnyzzz:
Fixes #221.

`CodeWrapperForCompilation`'s line mapping hardcoded a 12-default-import wrapped layout, but the default-import list has grown to 15 — so the compiler-error and stack-trace references agents see (`input.kt:N`) point 3 lines away from their actual code. The mapping's own tests pinned the hardcoded constants, so they stayed green while verifying the bug.

Fix: derive the offsets from `defaultImports.size` (user imports at `K+3`, user code at `K+11+N`), and rewrite the test expectations as computed `importLine(i)` / `codeLine(n, j)` helpers over the same layout — a future default-import addition can neither skew the mapping nor silently keep the tests green.

Verification: `:kotlin-cli:test` green (mapping + wrapper suites).

### DIFF
--- a/kotlin-cli/src/main/kotlin/com/jonnyzzz/mcpSteroid/koltinc/CodeWrapperForCompilation.kt
+++ b/kotlin-cli/src/main/kotlin/com/jonnyzzz/mcpSteroid/koltinc/CodeWrapperForCompilation.kt
@@ -21,6 +21,8 @@ object CodeWrapperForCompilation {
         "import com.intellij.openapi.editor.*",
         "import com.intellij.openapi.fileEditor.*",
         "import com.intellij.openapi.command.*",
+        // The type of McpScriptContext.progressIndicator (#213) — imported so fenced
+        // examples/scripts can name the type (e.g. a typed val) without an explicit import.
         "import com.intellij.psi.*",
         "import com.intellij.psi.search.*",
         "import com.intellij.psi.search.searches.*",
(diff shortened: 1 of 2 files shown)


jonnyzzz:
## Review 2026-08-03

- Superseded on main: PR #186 (merged 2026-07-04, c3dca9b8) rewrote the exact same two files. Main now builds the line mapping in lock-step with code emission via `WrappedLineEmitter` (`CodeWrapperForCompilation.kt`, `wrap()`), so there is no offset arithmetic left to drift — structurally stronger than this PR's derived `k+3` / `k+11+n` constants, which still encode the layout in two places (emitter + comment) and would break on any structural wrapper change, not just import-count changes.
- Main's implementation carries an explicit guard comment: "Never reintroduce magic offsets here." Merging this PR would reintroduce them.
- Issue #221 is CLOSED (2026-07-16) with the owner's verification comment naming this PR: "Open PR #231 is an obsolete earlier implementation and is not needed for this issue to be complete."
- Merge state is CONFLICTING/DIRTY — both touched files were fully rewritten by #186, so there is no salvageable delta.
- Minor: the added comment about `McpScriptContext.progressIndicator` (#213) sits above `import com.intellij.psi.*` and describes an import this PR does not add — it would have been misleading even pre-#186.

**Verdict: superseded** — #186 on main delivers the #221 fix with drift-proof tests; recommend closing this PR without merge.

PR #241, opened 2026-07-03
Survive the Gradle project-rename race in exec_code (x11k setup failures)

jonnyzzz:
First x11k CI runs: **11/14 green**; the 3 failures (993299640, 993296335, 993298537) shared one signature —

```
Process failed to register JDKs: ERROR: Project not found: "project-home".
Available project_name values: x-m90g7r9u
```

## Root cause
x11k is the first experiment repo whose Gradle `rootProject.name` differs from the checkout folder. When the Gradle sync lands, the IDE **renames the open project**, racing the driver's one-time name resolution in `mcpExecuteCode`'s default argument. 3 of 6 legs lost the race (timing-dependent); Keycloak/okhttp never hit it because their imports keep the folder-derived name.

## Fix
- `projectName` defaults to **null = resolve fresh on every retry attempt** (resolution is path-based, and paths survive renames).
- A `Project not found` result is a retry-signal **only when** the caller didn't pin a name **and** fresh resolution returns a *different* name (i.e. the rename actually landed). A pinned name or a stable re-resolution surfaces the error immediately — a genuinely missing project is never masked by an hour of polling (same fail-fast philosophy as the typed retry contract).
- TDD: `isProjectNotFound` regression test with the verbatim CI error text; `McpBusyRetryTest` 10/10.

After merge: re-trigger the 3 failed builds → expect the x11k matrix at 6/6.

### DIFF
--- a/test-integration/src/main/kotlin/com/jonnyzzz/mcpSteroid/integration/infra/mcp-steroid.kt
+++ b/test-integration/src/main/kotlin/com/jonnyzzz/mcpSteroid/integration/infra/mcp-steroid.kt
@@ -80,6 +80,18 @@ private const val INDEXING_POLL_BUDGET_MS = 60 * 60 * 1000L
 /** Pure: does this tool-result text say the IDE is still indexing (so we should call again)? */
 internal fun isIndexingInProgress(text: String): Boolean = text.contains(INDEXING_IN_PROGRESS_MARKER)
 
+/**
+ * Pure: does this tool-result text say the addressed project name does not exist (server-side)?
+ *
+ * On Gradle projects whose `rootProject.name` differs from the checkout folder, the IDE RENAMES the
+ * project when the Gradle sync lands (`project-home` → e.g. `x-m90g7r9u`) — racing the driver's
+ * name resolution. A request that resolved the name a moment too early gets this error even though
+ * the project is open and healthy (observed on the x11k builds 993299640/993296335/993298537). The
+ * caller re-resolves and retries; a name that no longer changes means it is NOT the rename race and
+ * the error is genuine.
+ */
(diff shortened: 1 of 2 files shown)


jonnyzzz:
## Review 2026-08-03

- **Superseded by main commit 73f448b2** ("test-integration: survive the Gradle import rename when addressing the project", merged 2026-08-03 via the #412 burn-down, merge 850ef0bf). It fixes the same rename race in `mcpExecuteCode` with the same shape: `projectName` defaults to null, re-resolve on the not-found answer, retry only when the fresh key differs.
- Main's version is strictly stronger on the points this PR left open: (a) `resolveProjectName()` now returns the `project_name` **routing key** from `steroid_list_projects` (path-matched), not the ambiguity-prone raw display name; (b) `isProjectNotFound(text, key)` matches the exactly-quoted key — this PR's bare `text.contains("Project not found:")` (mcp-steroid.kt:93) would let a script's own println masquerade as the routing failure and trigger a spurious hour-budget retry; (c) main also gates on `result.exitCode != 0` and resolves eagerly, so a genuinely-not-open project fails fast instead of paying a `list_projects` round-trip per poll attempt (this PR resolves fresh on every attempt, plus a second resolution inside the retry predicate).
- Test coverage on main (`ProjectRenameRetryTest`, 4 cases incl. different-key and script-chatter negatives) is a superset of the `McpBusyRetryTest` addition here.
- `mergeable=CONFLICTING` / `mergeStateStatus=DIRTY` — both sides rewrite the same `mcpExecuteCode` hunk in `test-integration/.../infra/mcp-steroid.kt`.
- Minor (moot): the PR commit carries an AI co-author trailer, which the repo's commit conventions ban.

**Verdict: superseded** — main already delivers this fix in a stronger form; recommend closing without merge.

PR #243, opened 2026-07-07
fix(devrig): dedup port-scanned IDEs already found via marker in devrig project

matveidetonator:
devrig project (and its text footer) listed every running IDE TWICE: once from its `.mcp-steroid` marker (plugin present, with projects) and again from the built-in web-server port scan, where the plugin isn't visible — so the same IDE showed both "MCP Steroid: <version>" and "MCP Steroid: not installed".

The two discoveries key on different ports of the same process (marker = plugin server + pid; port scan = built-in 63342/63343, no pid), so nothing correlated them. `renderProjectOutput3` passed the full port-scan set straight to the "skipped / not installed" footer.

Fix: before the footer, drop port IDEs whose normalised build matches any marker route — the same dedup `devrig backend` already does (normaliseBuildForDedup). Port IDEs with an unknown build are kept (cannot prove they are duplicates). Only the human/text path is affected; `--json` and the MCP tool use the marker-only handler and never had the footer.

Tests: fix the pre-existing render test's fixture (its port IDE accidentally shared the marker's build — that IS the bug) to model a genuinely plugin-less IDE, and add regression tests for the same-build dedup and the mixed case.

### DIFF
--- a/npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/ProjectCommand.kt
+++ b/npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/ProjectCommand.kt
@@ -53,7 +53,17 @@ fun renderProjectOutput3(
     val routesByIde: Map<DiscoveredIde, List<ProjectRoute>> = routes.groupBy { it.route }
     val reachableIdeCount = routesByIde.keys.size
 
-    if (routes.isEmpty() && portIdes.isEmpty()) {
+    // Drop port-discovered IDEs that a marker already represents (same normalised build): it is the
+    // SAME running IDE, seen a second time via its built-in web-server port, so listing it again as
+    // "MCP Steroid: not installed" is a duplicate. Mirrors the dedup `devrig backend` already does
+    // (see BackendCommand). Port IDEs with an unknown build are kept — we cannot prove they are dupes.
+    val markerBuilds = routes.mapNotNull { normaliseBuildForDedup(it.route.ide.build) }.toSet()
+    val skippedPortIdes = portIdes.filterTo(LinkedHashSet()) { portIde ->
+        val norm = normaliseBuildForDedup(portIde.buildNumber)
+        norm == null || norm !in markerBuilds
+    }
+
+    if (routes.isEmpty() && skippedPortIdes.isEmpty()) {
         out.println(NO_BACKENDS_DETECTED_MESSAGE)
         out.println()
         return
@@ -61,7 +71,7 @@ fun renderProjectOutput3(
 
     if (routes.isEmpty()) {
         out.println("No open projects across $reachableIdeCount backend(s).")
-        renderPortSkippedFooter(portIdes, out)
+        renderPortSkippedFooter(skippedPortIdes, out)
         out.println()
         return
     }
@@ -80,7 +90,7 @@ fun renderProjectOutput3(
         if (index < routes.lastIndex) out.println()
     }
 
-    renderPortSkippedFooter(portIdes, out)
+    renderPortSkippedFooter(skippedPortIdes, out)
     out.println()
 }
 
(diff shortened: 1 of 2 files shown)


npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/ProjectCommand.kt
```diff
@@ -53,15 +53,25 @@ fun renderProjectOutput3(
...
     val reachableIdeCount = routesByIde.keys.size
 
-    if (routes.isEmpty() && portIdes.isEmpty()) {
+    // Drop port-discovered IDEs that a marker already represents (same normalised build): it is the
+    // SAME running IDE, seen a second time via its built-in web-server port, so listing it again as
+    // "MCP Steroid: not installed" is a duplicate. Mirrors the dedup `devrig backend` already does
+    // (see BackendCommand). Port IDEs with an unknown build are kept — we cannot prove they are dupes.
+    val markerBuilds = routes.mapNotNull { normaliseBuildForDedup(it.route.ide.build) }.toSet()
+    val skippedPortIdes = portIdes.filterTo(LinkedHashSet()) { portIde ->
+        val norm = normaliseBuildForDedup(portIde.buildNumber)
```
jonnyzzz, line 62: There might be a more deeper bug -- could you please explain why we use buildNumber as the key here? 

That is totally accepted to have multiple copies of the same IDE build number, each operates with unique system/config/port number. 

We already have unique backend_id which can be used here as well. 

When multiple projects are open, there we can have multiple routes to the same IDE.

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/ProjectCommand.kt
```diff
@@ -53,15 +53,25 @@ fun renderProjectOutput3(
     val routesByIde: Map<DiscoveredIde, List<ProjectRoute>> = routes.groupBy { it.route }
     val reachableIdeCount = routesByIde.keys.size
 
-    if (routes.isEmpty() && portIdes.isEmpty()) {
+    // Drop port-discovered IDEs that a marker already represents (same normalised build): it is the
+    // SAME running IDE, seen a second time via its built-in web-server port, so listing it again as
+    // "MCP Steroid: not installed" is a duplicate. Mirrors the dedup `devrig backend` already does
+    // (see BackendCommand). Port IDEs with an unknown build are kept — we cannot prove they are dupes.
+    val markerBuilds = routes.mapNotNull { normaliseBuildForDedup(it.route.ide.build) }.toSet()
+    val skippedPortIdes = portIdes.filterTo(LinkedHashSet()) { portIde ->
```
jonnyzzz, line 61: `.filter { }` function defaults to LinkedHashSet, please check the docs

npx-kt/src/test/kotlin/com/jonnyzzz/mcpSteroid/devrig/ProjectCommandRenderTest.kt
```diff
@@ -167,15 +167,23 @@ class ProjectCommandRenderTest {
     // ----------------------------- skipped (port IDEs) ---------------------
 
     @Test
-    fun `port-only IDE appears in skipped footer and not in projects list`() {
-        val ide = markerIde(pid = 1L)
+    fun `port-only IDE (no matching marker) appears in skipped footer and not in projects list`() {
```
jonnyzzz, line 170: I would add a test for the list_projects tool, just to make sure it calls the same code and is not suddenly changed

jonnyzzz (changes requested):

jonnyzzz:
@matveidetonator what is the status of that PR?

matveidetonator:
@jonnyzzz thanks for the review. The build-keyed dedup fixes the symptom but is fragile for the "multiple copies of the same build" case you flagged, so I'm closing this PR and tracking the problem properly in #263.

#263 captures: what the two discovery paths actually share (only `build` today — `backend_id` is keyed differently on each path, so it can't correlate marker↔port), and the options (A: write the built-in port into the marker for an exact process-level match; C: a minimal build-dedup-only-when-unambiguous fix for now), plus extracting one shared identity helper for both `project` and `backend`.

Draft fix preserved on branch `fix/project-list-dedup-port-footer`.

Issue #266, opened 2026-07-10
devrig CLI: make --project_name optional — infer the project from the current working directory


matveidetonator:
Epic: #188 (MCP as CLI). 

### Motivation

Today the project-scoped CLI commands (`execute_code`, `take_screenshot`, `input`, `execute_feedback`) require an explicit `--project_name`. For an agent that's an extra mandatory step: it must first call `devrig project`/`list_projects`, parse the output, and substitute the name. That exact step is what broke the MCP arm in #251 (a hard-coded `project-home`). The natural default: **if a command is run from a directory inside a project, infer the project automatically.**

### Idea

devrig already has a snapshot mechanism that collects all backends and open projects and correlates them: `projectRouting.routes()` → `List<ProjectRoute>`, where each route carries `projectPath`, `exposedProjectName`, and `route` (the backend). See `ProjectCommand.kt` / `renderProjectJson3`.

Algorithm for an optional `--project_name`:

1. Take the cwd (the directory the command was launched from), normalize to an absolute path.
2. From the `routes()` snapshot, find the route whose `projectPath` is the **longest prefix** of the cwd (i.e. cwd lives inside the project directory); with nested projects, the longest (most specific) match wins.
3. Exactly one candidate → use it silently (optionally print to stderr what was chosen and why).
4. No match, or several at the same level → **fail fast** with an agent-understandable error: list the candidates and ask for an explicit `--project_name` (in the spirit of the #188 contract — usage errors must be agent-understandable).

### Scope

- A shared "cwd → project_name" resolver reused by all project-scoped commands (`execute_code`, `take_screenshot`, `input`, `execute_feedback`).
- `--project_name` becomes optional; an explicitly passed value always overrides inference.
- Longest-prefix logic with path-segment boundaries (so `/foo/bar` does not "match" `/foo/barbaz`).
- Unit tests for the resolver: single project, nested projects, zero matches, ambiguity, cwd == project root.
- Update the commands' `--help`/docs: describe the default behavior and when the name is still required.

### Acceptance criteria

- `cd <project-dir> && devrig execute_code --code-file=x.kts` works without `--project_name`.
- From a directory outside any project — a clear error listing the open projects.
- An explicit `--project_name` still works and takes precedence.

### Note

This directly removes the class of errors from #251 for the CLI arm and drops the mandatory "first find out the project name" step for the agent.

PR #272, opened 2026-07-14
Cli hardening

matveidetonator:
every mcp tool now is available as devrig command, layered help execute_code
tests: contract, parse, error-envelope, Docker smoke test

### DIFF
--- a/mcp-steroid-server/src/main/kotlin/com/jonnyzzz/mcpSteroid/server/FetchResourceToolHandler.kt
+++ b/mcp-steroid-server/src/main/kotlin/com/jonnyzzz/mcpSteroid/server/FetchResourceToolHandler.kt
@@ -18,6 +18,7 @@ import com.jonnyzzz.mcpSteroid.prompts.generated.prompt.DebuggerSkillPromptArtic
 import com.jonnyzzz.mcpSteroid.prompts.generated.prompt.SkillPromptArticle
 import com.jonnyzzz.mcpSteroid.prompts.generated.prompt.TestSkillPromptArticle
 import com.jonnyzzz.mcpSteroid.prompts.generated.skill.CodingWithIntelliJPromptArticle
+import com.jonnyzzz.mcpSteroid.prompts.generated.skill.ExecuteCodeToolDescriptionPromptArticle
 import com.jonnyzzz.mcpSteroid.thisLogger
 
 /**
@@ -106,4 +107,17 @@ fun canonicalResourceEntryPoints(): List<String> = listOf(
     FindDuplicatesPromptArticle().uri,
     InspectAndFixPromptArticle().uri,
     CodingWithIntelliJPromptArticle().uri,
+)
+
+/**
+ * Layered "go deeper" guide URIs for the `devrig execute_code` CLI help — ordered shallow→deep so an
+ * agent can fetch only as far as it needs (`devrig prompt <uri>`). Built from the generated article
+ * classes (no hardcoded `mcp-steroid://` literals), so a renamed/removed article breaks the build.
+ * Each pair is `uri to one-line-what-you-get`.
+ */
+fun executeCodeGuideUris(): List<Pair<String, String>> = listOf(
+    ExecuteCodeToolDescriptionPromptArticle().uri to "full tool guide: decision tree, threading rules, multi-file edits, output rules",
+    CodingWithIntelliJPromptArticle().uri to "IntelliJ API reference: PSI, VFS, refactorings, patterns, examples",
+    TestSkillPromptArticle().uri to "running/inspecting tests via the IDE runner",
+    DebuggerSkillPromptArticle().uri to "breakpoints, debug sessions, threads",
 )
--- a/mcp-steroid-server/src/main/kotlin/com/jonnyzzz/mcpSteroid/server/FetchResourceToolHandler.kt
+++ b/mcp-steroid-server/src/main/kotlin/com/jonnyzzz/mcpSteroid/server/FetchResourceToolHandler.kt
@@ -10,6 +10,7 @@ import com.jonnyzzz.mcpSteroid.mcp.get
 import com.jonnyzzz.mcpSteroid.mcp.param
 import com.jonnyzzz.mcpSteroid.mcp.required
 import com.jonnyzzz.mcpSteroid.mcp.string
+import com.jonnyzzz.mcpSteroid.prompts.PromptsContext
 import com.jonnyzzz.mcpSteroid.prompts.generated.ResourcesIndex
 import com.jonnyzzz.mcpSteroid.prompts.generated.ide.FindDuplicatesPromptArticle
 import com.jonnyzzz.mcpSteroid.prompts.generated.ide.InspectAndFixPromptArticle
@@ -66,15 +67,43 @@ class FetchResourceToolHandler(
         log.info("steroid_fetch_resource: $uri")
 
         val promptsContext = handler().buildPromptsContext(projectName)
-        val article = ResourcesIndex().roots.values
-            .asSequence()
-            .flatMap { it.articles.values.asSequence() }
-            .firstOrNull { it.uri == uri && it.filter.matches(promptsContext) }
+        val payload = resolveResourcePayload(uri, promptsContext)
             ?: return ToolCallResult(
                 content = listOf(ContentItem.Text(text = "ERROR: Resource not found: $uri")),
                 isError = true
             )
 
-        return ToolCallResult(content = listOf(ContentItem.Text(text = article.readPayload(promptsContext))))
+        return ToolCallResult(content = listOf(ContentItem.Text(text = payload)))
     }
 }
+
+/**
+ * Resolves a `mcp-steroid://` resource URI to its rendered Markdown payload for the given
+ * [context], or `null` when no article matches the URI + IDE filter.
+ *
+ * The single source of truth for URI → payload resolution: used by [FetchResourceToolHandler]
+ * (the `steroid_fetch_resource` MCP tool) and by the `devrig fetch_resource` / `devrig prompt`
+ * CLI commands, so both surfaces resolve identically.
+ */
+fun resolveResourcePayload(uri: String, context: PromptsContext): String? {
+    val article = ResourcesIndex().roots.values
+        .asSequence()
+        .flatMap { it.articles.values.asSequence() }
+        .firstOrNull { it.uri == uri && it.filter.matches(context) }
+        ?: return null
+    return article.readPayload(context)
+}
+
+/**
+ * Canonical entry-point URIs to suggest when a fetch misses — built from the generated article
+ * classes (never hardcoded `mcp-steroid://` literals, per NoHardcodedMcpSteroidUriUsageTest).
+ * Reused by CLI error hints so the suggestions stay in sync with the tool description above.
+ */
+fun canonicalResourceEntryPoints(): List<String> = listOf(
+    SkillPromptArticle().uri,
+    TestSkillPromptArticle().uri,
+    DebuggerSkillPromptArticle().uri,
+    FindDuplicatesPromptArticle().uri,
+    InspectAndFixPromptArticle().uri,
+    CodingWithIntelliJPromptArticle().uri,
+)
--- a/mcp-steroid-server/src/main/kotlin/com/jonnyzzz/mcpSteroid/server/FetchResourceToolHandler.kt
+++ b/mcp-steroid-server/src/main/kotlin/com/jonnyzzz/mcpSteroid/server/FetchResourceToolHandler.kt
@@ -10,6 +10,7 @@ import com.jonnyzzz.mcpSteroid.mcp.get
 import com.jonnyzzz.mcpSteroid.mcp.param
 import com.jonnyzzz.mcpSteroid.mcp.required
 import com.jonnyzzz.mcpSteroid.mcp.string
+import com.jonnyzzz.mcpSteroid.prompts.ArticleBase
 import com.jonnyzzz.mcpSteroid.prompts.PromptsContext
 import com.jonnyzzz.mcpSteroid.prompts.generated.ResourcesIndex
 import com.jonnyzzz.mcpSteroid.prompts.generated.ide.FindDuplicatesPromptArticle
@@ -68,45 +69,43 @@ class FetchResourceToolHandler(
         log.info("steroid_fetch_resource: $uri")
 
         val promptsContext = handler().buildPromptsContext(projectName)
-        val payload = resolveResourcePayload(uri, promptsContext)
+        val article = resolveResourceArticle(uri, promptsContext)
             ?: return ToolCallResult(
                 content = listOf(ContentItem.Text(text = "ERROR: Resource not found: $uri")),
                 isError = true
             )
 
-        return ToolCallResult(content = listOf(ContentItem.Text(text = payload)))
+        return ToolCallResult(content = listOf(ContentItem.Text(text = article.readPayload(promptsContext))))
     }
 }
 
 /**
- * Resolves a `mcp-steroid://` resource URI to its rendered Markdown payload for the given
- * [context], or `null` when no article matches the URI + IDE filter.
+ * Resolves a `mcp-steroid://` resource URI to its matching [ArticleBase] for the given [context],
+ * or `null` when no article matches the URI + IDE filter. Callers render the payload themselves via
+ * [ArticleBase.readPayload] — this function hands back the article object, not pre-rendered text.
  *
- * The single source of truth for URI → payload resolution: used by [FetchResourceToolHandler]
+ * The single source of truth for URI → article resolution: used by [FetchResourceToolHandler]
  * (the `steroid_fetch_resource` MCP tool) and by the `devrig fetch_resource` / `devrig prompt`
  * CLI commands, so both surfaces resolve identically.
  */
-fun resolveResourcePayload(uri: String, context: PromptsContext): String? {
-    val article = ResourcesIndex().roots.values
+fun resolveResourceArticle(uri: String, context: PromptsContext): ArticleBase? =
+    ResourcesIndex().roots.values
         .asSequence()
         .flatMap { it.articles.values.asSequence() }
         .firstOrNull { it.uri == uri && it.filter.matches(context) }
-        ?: return null
-    return article.readPayload(context)
-}
 
 /**
- * Canonical entry-point URIs to suggest when a fetch misses — built from the generated article
+ * Canonical entry-point articles to suggest when a fetch misses — built from the generated article
  * classes (never hardcoded `mcp-steroid://` literals, per NoHardcodedMcpSteroidUriUsageTest).
  * Reused by CLI error hints so the suggestions stay in sync with the tool description above.
  */
-fun canonicalResourceEntryPoints(): List<String> = listOf(
-    SkillPromptArticle().uri,
-    TestSkillPromptArticle().uri,
-    DebuggerSkillPromptArticle().uri,
-    FindDuplicatesPromptArticle().uri,
-    InspectAndFixPromptArticle().uri,
-    CodingWithIntelliJPromptArticle().uri,
+fun canonicalResourceEntryPoints(): List<ArticleBase> = listOf(
+    SkillPromptArticle(),
+    TestSkillPromptArticle(),
+    DebuggerSkillPromptArticle(),
+    FindDuplicatesPromptArticle(),
+    InspectAndFixPromptArticle(),
+    CodingWithIntelliJPromptArticle(),
 )
 
 /**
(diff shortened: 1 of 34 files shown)


mcp-steroid-server/src/main/kotlin/com/jonnyzzz/mcpSteroid/server/FetchResourceToolHandler.kt
```diff
@@ -66,15 +68,56 @@ class FetchResourceToolHandler(
...
+ * (the `steroid_fetch_resource` MCP tool) and by the `devrig fetch_resource` / `devrig prompt`
+ * CLI commands, so both surfaces resolve identically.
+ */
+fun resolveResourcePayload(uri: String, context: PromptsContext): String? {
+    val article = ResourcesIndex().roots.values
+        .asSequence()
+        .flatMap { it.articles.values.asSequence() }
+        .firstOrNull { it.uri == uri && it.filter.matches(context) }
+        ?: return null
+    return article.readPayload(context)
```
jonnyzzz: Everything is right, but out of the context, I would return Article here, not the plain text.

mcp-steroid-server/src/main/kotlin/com/jonnyzzz/mcpSteroid/server/FetchResourceToolHandler.kt
```diff
@@ -66,15 +68,56 @@ class FetchResourceToolHandler(
...
+}
+
+/**
+ * Canonical entry-point URIs to suggest when a fetch misses — built from the generated article
+ * classes (never hardcoded `mcp-steroid://` literals, per NoHardcodedMcpSteroidUriUsageTest).
+ * Reused by CLI error hints so the suggestions stay in sync with the tool description above.
+ */
+fun canonicalResourceEntryPoints(): List<String> = listOf(
```
jonnyzzz: you may return Articles, to avoid bare strings

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/CliToolSupport.kt
```diff
@@ -0,0 +1,249 @@
...
+    if (json) {
+        out.println(toEnvelopeJson(command))
+        return if (isError) CliExit.TOOL_ERROR else CliExit.OK
+    }
+
+    val sink = if (isError) err else out
+    for (item in content) {
+        when (item) {
+            is ContentItem.Text -> sink.println(item.text)
+            is ContentItem.Image -> sink.println("[image: ${item.mimeType}, ${item.decodedByteCount()} bytes]")
```
jonnyzzz: you either need to encode as data-url or return a path to a file, somewhere under the ~/.mcp-steroid/ temp folder.

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/CliToolSupport.kt
```diff
@@ -0,0 +1,249 @@
...
+ * Contract (locked by tests, matches the rest of devrig):
+ *  - Non-error text content → **stdout** (so `| jq`, `| less` work).
+ *  - Error content (`isError == true`) → **stderr**; stdout stays empty.
+ *  - `--json` emits a single stable envelope on stdout regardless of success/failure.
+ *
+ * @param command the CLI subcommand name, echoed into the JSON envelope for context.
+ */
+fun ToolCallResult.renderTo(
+    command: String,
+    json: Boolean,
```
jonnyzzz: json: Boolean  <-- here in this file look like a missing abstraction, you may have presentation, which is implemented different -- one for --json mode and the other for console more. It may let you deal with more high-level code in the rest of the impl

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/CliToolSupport.kt
```diff
@@ -0,0 +1,249 @@
...
+ * The single, unified `--json` envelope shared by EVERY `devrig` MCP-as-CLI command:
+ *
+ * ```json
+ * { "tool": {"name":"devrig","version":"..."}, "command": "<name>", "isError": <bool>, "data": { ... } }
+ * ```
+ *
+ * `data` holds the command-specific payload: `{content:[...]}` for tool-result commands (fetch_resource,
+ * execute_code, …), `{projects:[...]}` for list_projects, `{windows,backgroundTasks}` for list_windows.
+ * One outer shape keeps agents and tests from special-casing per command. The [Json] instance is shared
+ * so encoding stays consistent.
```
jonnyzzz: probably too detailed comments here and everywhere, I see it's `prettyPrint=true`. I would add new-lines to the `Json { \n ... }` and avoid `;`s

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/CliToolSupport.kt
```diff
@@ -0,0 +1,249 @@
...
+ */
+fun ToolCallResult.toEnvelopeJson(command: String): String =
+    cliEnvelopeJson(command, isError, contentDataJson())
+
+/**
+ * Builds the `data:{content:[...]}` object for a [ToolCallResult] — the same payload
+ * [toEnvelopeJson] wraps, but exposed so a command can merge command-specific keys alongside it.
+ */
+fun ToolCallResult.contentDataJson(): JsonObject = buildJsonObject {
+    putJsonArray("content") {
```
jonnyzzz: IIRC, ToolCallResult is already `@Serializable`, you may call it's native serializer instead. Questionable why we need that (and it's the second copy of the logic in that file). Not yet clear to me

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/CliToolSupport.kt
```diff
@@ -0,0 +1,249 @@
...
+        for (item in content) {
+            add(buildJsonObject {
+                when (item) {
+                    is ContentItem.Text -> {
+                        put("type", "text")
+                        put("text", item.text)
+                    }
+                    is ContentItem.Image -> {
```
jonnyzzz: I would keep data as-is, since it's sent from the client (alternative, save to files, as suggested above). The image file size will not help a caller agent to proceed

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/DevrigBeacon.kt
```diff
@@ -72,6 +72,14 @@ class DevrigBeacon(
             is DevrigCommand.DevrigCommandProject -> "project"
             is DevrigCommand.DevrigCommandInstall -> "install"
             is DevrigCommand.DevrigCommandInstallDevrig -> "install"
+            is DevrigCommand.DevrigCommandFetchResource -> "fetch_resource"
+            is DevrigCommand.DevrigCommandExecuteCode -> "execute_code"
+            is DevrigCommand.DevrigCommandListProjects -> "list_projects"
+            is DevrigCommand.DevrigCommandListWindows -> "list_windows"
+            is DevrigCommand.DevrigCommandOpenProject -> "open_project"
```
jonnyzzz, line 79: task for the next iteration and for the next PR: think how you keep that information in the ToolSpec classes, instead of hardcoding all of that here? Ideal scenario -- I can add a new MCP tool which will automatically (or at minimal cost) be registered to the CLI as well. How would you design it?

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/FetchResourceCommand.kt
```diff
@@ -0,0 +1,83 @@
...
+ * `devrig prompt <uri>` / `devrig fetch_resource --uri=...` — the CLI face of `steroid_fetch_resource`.
+ *
+ * Bundled `mcp-steroid://` articles ship inside the devrig binary, so this resolves **without a
+ * running IDE** using [PromptsContext.Generic]. Passing `--project_name` upgrades to that project's
+ * IDE-specific context via the existing routing + [DevrigPromptsContextHandler] (which needs the IDE
+ * to be discovered). URI → payload resolution is the shared [resolveResourcePayload] — the same code
+ * the MCP tool uses, so both surfaces render identically.
+ */
+fun DevrigServices.runFetchResourceCommand(command: DevrigCommand.DevrigCommandFetchResource): Int {
+    val uri = command.uri
```
jonnyzzz, line 29: I would use `FetchResourceToolHandler` for that to avoid duplication of the logic. All we need is to map parameters from the CLI to the tool call JSON parameters. Potentially a small abstraction is needed. 

Prompts is a bit more special tool call, so it might be implemented differently, but for the majority of commands that makes a log of sense -- we should avoid code duplication (it will be hard to maintain to handle the same bugs twice)

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/FetchResourceCommand.kt
```diff
@@ -0,0 +1,83 @@
...
+private fun DevrigServices.resolvePromptsContext(projectName: String?): PromptsContext {
+    if (projectName.isNullOrBlank()) return PromptsContext.Generic
+    val route = try {
+        projectRouting.requireProject(projectName)
+    } catch (e: IllegalArgumentException) {
+        throw PromptsContextResolutionException(
+            "unknown --project_name '$projectName' (get it from `devrig list_projects`)"
+        )
+    }
+    return DevrigPromptsContextHandler.promptsContextFromBuild(route.route.ide.build)
```
jonnyzzz: I think we need to refactor (maybe as a small PR before) to have the function to create Prompts Context from route directly (and thus we hide the `promptsContextFromBuild(route.route.ide.build)` detail. Rebase this PR if you'd follow that way

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/HelpCommand.kt
```diff
@@ -1,12 +1,66 @@
...
     return 0
 }
 
+/**
+ * Layered help: `devrig help <topic>` / `devrig <cmd> --help`. A [topic] gives a concise per-command
+ * entry that ends with `devrig prompt <uri>` pointers so an agent drills only as deep as it needs;
+ * an unknown/absent topic falls back to the global banner.
+ */
+fun printTopicHelp(topic: String?, out: PrintStream): Int = when (topic) {
+    "execute_code" -> printExecuteCodeHelp(out)
```
jonnyzzz, line 17: why only one command? 

I would more that help details to each `FetchResourceToolHandler`, to avoid consolidating all hints here. 

The command parameters are easy to generate from the MCP Tool specification (you can expand the `InputSchemaElement` hierarchy for that if needed)

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/HelpCommand.kt
```diff
@@ -59,6 +113,40 @@ fun printHelp(out: PrintStream) : Int {
...
                                          print manual MCP Steroid plugin install
                                          instructions for that IDE.
+
+        MCP tools as CLI (same tools as the `devrig mcp` server, callable from the shell):
+
+          devrig prompt <uri> [--project_name <key>]
+                                         fetch a mcp-steroid:// guide by URI (steroid_fetch_resource).
+                                         Works without a running IDE for bundled docs; pass
+                                         --project_name for IDE-specific content.
+          devrig fetch_resource --uri=<uri> [--project_name <key>]
```
jonnyzzz, line 126: in this section -- let's remove duplication, I think we need to re-used information from the MCP tools to generate these texts

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/Main.kt
```diff
@@ -85,12 +85,18 @@ suspend fun DevrigServices.mainImpl2(
...
+    // The devrig binary owns ~/.mcp-steroid/bin/devrig: (re)create/update it on start so it self-heals
+    // and always points at this running install + JDK. Best-effort and stderr-only — never blocks
+    // serving. It writes atomically, so an agent mid-read of the launcher never sees a torn file. For
+    // `devrig mcp` we skip the Windows user-PATH registration (it spawns PowerShell and would delay the
+    // first serve); that runs on interactive/`install` invocations instead.
+    //
+    // Gated on [selfHealsLauncherOnStart]: the MCP-as-CLI tool facades are thin, stateless bridge
+    // forwarders and must NOT mutate on-disk launcher/PATH state (Tenet 3). The self-heal is a
+    // bootstrap/lifecycle concern reserved for `mcp` + the interactive commands.
+    if (command.selfHealsLauncherOnStart()) {
```
jonnyzzz, line 97: this change looks like outside of that PR. Should we fork it to another PR and rebase this one?

jonnyzzz: We also have several related issues to that logic already

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/ToolBackedCommands.kt
```diff
@@ -0,0 +1,386 @@
...
+ * wiring the `devrig mcp` stdio proxy uses, so the CLI never reimplements tool logic. Each command takes
+ * `tools` as a defaulted parameter purely so tests can inject a fake and assert the args→`*Params`→render
+ * glue without a live IDE (payload→wire mapping is covered by DevrigToolBridgeClientTest).
+ */
+
+/**
+ * Runs [block] against [tools], turning routing/bridge failures into meaningful exit codes + agent-usable
+ * stderr messages, then renders the [ToolCallResult].
+ */
+private inline fun DevrigServices.runToolCall(
```
jonnyzzz, line 49: it feels like we already have all the implementation to run tool calls, that is how `devrig mcp` works. Feels like a duplication of logic so far

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/ToolBackedCommands.kt
```diff
@@ -0,0 +1,386 @@
...
+ * file, non-regular file, malformed path string) from a genuine read failure ([CliExit.IO_ERROR] — the
+ * path resolves and is a regular file but reading it throws, e.g. a permission denial).
+ */
+private class CodeArgException(message: String, val exit: Int) : RuntimeException(message)
+
+/** Reads inline `--code` or the `--code-file` path; throws [CodeArgException] on a bad/unreadable file. */
+private fun resolveCodeArg(inline: String?, file: String?): String {
+    if (!inline.isNullOrBlank()) return inline
+    val path = try {
+        Path.of(file!!)
```
jonnyzzz: handle `!!` more nicely

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/ToolBackedCommands.kt
```diff
@@ -0,0 +1,386 @@
...
+// ----------------------------------- execute_code -----------------------------------
+
+fun DevrigServices.runExecuteCodeCommand(
+    command: DevrigCommand.DevrigCommandExecuteCode,
+    tools: McpSteroidTools = StubMcpSteroidTools(this),
+): Int {
+    // `--code-file=-` reads the script from stdin so agents can pipe a snippet without a temp file.
+    val code = try {
+        when {
+            command.codeFile == "-" -> mcpStdin.readBytes().decodeToString()
```
jonnyzzz, line 165: nice feature, but it was not documented above.
How does readBytes() work if remove client sends data slowly? Will it keep reading or stop?

jonnyzzz (changes requested):

matveidetonator:
implemented:
  - replaced threaded json boolean with presentation
  - reused ToolSpec.call() in cli commands
  - fixed image rendering and native json serialization
  - implemented issue #266 
  - fixed small review comments

regarding c11 and c12, the decision to do only execute_code with extended help was conscious: its prompt is significantly larger, while the other commands have short descriptions

kept self heal gate in this PR because the new cli commands need to remain stateless (c13, c14)

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/DevrigBeacon.kt
matveidetonator: solved in #351

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/FetchResourceCommand.kt
matveidetonator: solved in #351

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/FetchResourceCommand.kt
matveidetonator: solved in #351

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/HelpCommand.kt
matveidetonator: solved in #351

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/HelpCommand.kt
matveidetonator: solved in #351

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/Main.kt
matveidetonator: solved in #351

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/ToolBackedCommands.kt
matveidetonator: solved in #351

jonnyzzz:
## Review 2026-08-03

- **Staleness (primary finding):** `mergeable=CONFLICTING`. Verified via `git merge-tree`: content conflicts in `npx-kt/.../Cli.kt`, `Main.kt`, `DevrigBeacon.kt`. The branch's merge-base is 159 commits behind main.
- **Semantic rebase hazard #1 — #356:** main (2026-07-31) now has `CliToolSpec` + `McpSteroidTools.commonToolSpecs()/devrigToolSpecs()`, extracted from this track's phase B. This branch's "route CLI tool calls through ToolSpec.call()" commits (4b10b2c1, 8b0f5764) target the pre-#356 spec shape and must be reconciled onto the merged one. Afterwards #351 (stacked on this branch) needs re-stacking too — its foundation commits are already on main.
- **Semantic rebase hazard #2 — #397/#399:** main's `runCli` now centrally handles `AgentCliNotLaunchableException`; this PR's `--json` envelope catch-all in `Main.kt` `runCli` overlaps it. Main also calls `ensureBinLauncher` unconditionally on start, while this PR gates it via `selfHealsLauncherOnStart()` (Main.kt:97) — the Tenet-3 gate changes the every-start bin-launcher self-heal policy and needs explicit owner sign-off during rebase.
- **Review-response state:** the 2026-07-14 CHANGES_REQUESTED items (C1–C17) were addressed in the a40bf1e2..8553cf3f series; three threads (r3579479002, r3579494087, r3579514905) are deliberately deferred to #351 per its description. Owner re-review pending since 2026-07-17.
- **Not superseded, spot-checks clean:** main has none of the MCP-as-CLI command files (`ToolBackedCommands.kt`, `CliToolSupport.kt`, etc.) and #351 builds on this branch rather than replacing it. No banned patterns found in the diff; test coverage is credible (parse/contract/envelope glue tests + opt-in Docker smoke `CliDevrigToolsIntegrationTest`, whose infra — `DevrigSteroidDriver`, `waitForProjectReady` — survived the #412 test-infra rework).

**Verdict: needs-rebase** — content is sound and the requested changes were addressed, but the base is stale/conflicting with non-trivial semantic reconciliation against #356/#397; rebase onto main (dropping the now-redundant pre-#356 ToolSpec adaptation), then re-stack #351.

Issue #274, opened 2026-07-14
install.ps1: Ineffective download + incorrect progress


jonnyzzz:
The download flow looks not too fast. The progress indication is mis-positioned on the screen. 
It feels like it slows down the progress, because it updates the screen for every byte

<img width="2111" height="1641" alt="Image" src="https://github.com/user-attachments/assets/6f4bb75d-bd20-47e7-a1f0-35bae01474cd" />

jonnyzzz:
The progress is the default progress of powershell,
see https://github.com/PowerShell/PowerShell/issues/2138

Unpack archive and download performance are affected with the bug, 
the recommended workaround is to disable progress, e.g. as of 
https://stackoverflow.com/questions/18770723/hide-progress-of-invoke-webrequest

jonnyzzz:
Fixed on main by 1276d54d + bde30226: install.ps1 now silences PowerShell's default progress (the per-byte redraw bug, PowerShell/PowerShell#2138) and restores $ProgressPreference in a finally block. The replacement visible bounded progress line ships with PR #363 under #228. Closing; note the live devrig.dev script picks this up when the website branch next advances.

jonnyzzz:
Fixed in **v0.102**. Release notes: https://devrig.dev/releases/0.102/ · GitHub release: https://github.com/jonnyzzz/mcp-steroid/releases/tag/v0.102

PR #281, opened 2026-07-15
install.ps1: fix Windows PowerShell 5.1 arch detect + progress (fix #273 #274)

### LINKED ISSUE #273: install.ps1: OSArchitecture is not set
OSArchitecture is not set, so the script fails. 

In addition to that, we need to double check we close stdin for the processes which we start

Image

Image

jonnyzzz:
Fixes the two crashes/regressions that break `install.ps1` on the default Windows 10/11 shell (**Windows PowerShell 5.1**) during `irm | iex` bootstrap.

## #274 — progress bar makes the install crawl
PowerShell's default progress bar redraws on every byte read (`Invoke-WebRequest`) and every entry unpacked (`Expand-Archive`), turning the multi-hundred-MB download + JDK unpack into a UI-bound crawl. → `$ProgressPreference = 'SilentlyContinue'` for the whole script (status still goes to stderr via `Write-Log`).

## #273 — arch detection aborts under StrictMode
Detection read `[RuntimeInformation]::OSArchitecture` as the **primary** lookup. That property is missing on the .NET Framework 4.6.x base Windows 10 ships, and `Set-StrictMode -Version Latest` turns the miss into a hard `PropertyNotFoundStrict` abort (`OSArchitecture cannot be found`) — the whole install dead mid-bootstrap. → detection now reads the always-present Windows env vars `$env:PROCESSOR_ARCHITEW6432` (WoW64 32-on-64 first) / `$env:PROCESSOR_ARCHITECTURE`; `[RuntimeInformation]` is only a non-Windows (pwsh Core) fallback, **guarded by try/catch** so no future strict-mode tightening can re-introduce #273.

## Coverage
- Unit tests pin `PROCESSOR_ARCHITECTURE` as the primary path and gate any `RuntimeInformation` access behind a preceding `try{}`.
- `installerIntegrationTest` drops `DEVRIG_OS`/`DEVRIG_CPU` so the pwsh-on-Ubuntu Docker run exercises the auto-detect path.
- New `InstallerPs1ExecutionTest` runs the generated `install.ps1` end-to-end under native `powershell.exe` (PS 5.1) + `pwsh`, regression-guarding the exact #273 stderr signatures.
- A throwaway `installer-ps1-verify.yml` reproduces both lanes on GH CI (permanent home: TeamCity).

## Split out of this PR
- **stdin-close contract** (`< /dev/null` / `$null |` on the `devrig install devrig` handoff) → moved to its own foundational PR #282.
- **Windows shell-form sh-script hook regression** (`ClaudeAgentLaunchTest.hook-devrig-pattern`) → real Claude-on-Windows regression, tracked in #283. The verify workflow's Windows lane runs the full suite with `--continue`, so that one case shows an honest red without gating this fix; our `install.ps1` tests pass and are visible in the report.

### DIFF
--- a/.github/workflows/installer-ps1-verify.yml
+++ b/.github/workflows/installer-ps1-verify.yml
@@ -1,0 +1,73 @@
+# Cross-OS validation for the generated installers (install.sh / install.ps1). Runs on EVERY branch
+# push so a template edit that would break the `curl | sh` / `irm | iex` bootstrap on a real OS is
+# caught before it lands — this is the check that would have caught jonnyzzz/mcp-steroid#273/#274.
+# (TeamCity is the internal home for the same suites; this GH lane keeps the public fork covered too.)
+#
+# Two lanes:
+#   * ubuntu-latest → :installer-gen:installerIntegrationTest (Linux Docker + pwsh 7.x)
+#   * windows-latest → :test-integration-agent-launch:test (Windows PowerShell 5.1 + pwsh)
+#
+# The Windows lane runs InstallerPs1ExecutionTest end-to-end under powershell.exe (the default shell
+# on Windows 10/11). Neither lane sets DEVRIG_OS / DEVRIG_CPU: both exercise the full auto-detect path.
+name: Installer PS1 Verify
+
+on:
+  push:
+    branches: ['**']
+  workflow_dispatch:
+
+concurrency:
+  # One run per ref; a newer push cancels the in-flight outdated run.
+  group: installer-ps1-verify-${{ github.ref }}
+  cancel-in-progress: true
+
+# Opt JavaScript actions onto Node 24 to match the repo's other workflows.
+env:
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: "true"
+
+jobs:
+  linux-docker:
+    name: linux-docker (pwsh 7.x)
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-java@v4
+        with: { distribution: temurin, java-version: '25' }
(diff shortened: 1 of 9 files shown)


jonnyzzz:
### Update: codex review addressed + coverage hardened

Ran a codex peer review (via `run-agent.sh`) over both this PR and the stdin PR, then acted on the findings:

**Correctness**
- Fixed a `$ProgressPreference` **session leak**: `irm | iex` runs the script in the caller's scope, so the bare `SilentlyContinue` assignment permanently silenced progress in the user's session. Now snapshot + restore in a `finally`. Verified in a `pwsh` container that the caller's value is restored after an uncaught error.
- `$null |` / `< /dev/null` / arch regex confirmed correct by codex; the misleading PS rationale wording was corrected in #285 (follow-up to the merged #282).

**Convention violations fixed**
- Replaced the banned runtime skips (`assumeTrue(isWindows)` / `assumeTrue(pwshOnPath)`) in `InstallerPs1ExecutionTest` with a **task-level** Windows gate: new `windowsPs1Test` task (`enabled = isWindows`); the cross-OS `test` task excludes that class (still reported under its own task). pwsh is now in-test extra coverage, not a skip. The `isCommandOnPath` catch logs instead of swallowing.

**New runtime coverage** (all validated locally under `mcr.microsoft.com/powershell`, closing the top blind spots):
- Arch auto-detect matrix — WoW64 (`ARCHITEW6432` beats `x86`), ARM64, x86-rejection.
- Negative path — corrupt devrig sha → `SHA-256 mismatch`.
- Behavioral stdin close — fake devrig reads stdin and records immediate EOF.
- ProgressPreference restore — dot-sourced run confirms no session leak.

The intermittent `ClaudeAgentLaunchTest.hook-devrig-pattern` remains tracked separately in #283 (test-harness timing race, not a product bug).

Issue #150, opened 2026-06-24
devrig hangs on Windows: ensureWindowsPathEntry spawns PowerShell with open stdin + no output capture (waitFor blocks)


jonnyzzz:
## Symptom
On Windows, `devrig` start/install can hang during user-PATH registration. A thread dump shows a `DefaultDispatcher` worker stuck in `Process.waitFor` for ~113s (RUNNABLE, never returns):

```
"DefaultDispatcher-worker-1" #41 daemon ... elapsed=113.16s ... runnable
   java.lang.Thread.State: RUNNABLE
        at java.lang.ProcessImpl.waitForInterruptibly(java.base@25.0.3/Native Method)
        at java.lang.ProcessImpl.waitFor(java.base@25.0.3/ProcessImpl.java:555)
        at com.jonnyzzz.mcpSteroid.devrig.BinLauncherKt.ensureWindowsPathEntry(BinLauncher.kt:312)
        at com.jonnyzzz.mcpSteroid.devrig.BinLauncherKt.ensureBinLauncherCore(BinLauncher.kt:149)
        at com.jonnyzzz.mcpSteroid.devrig.BinLauncherKt.ensureBinLauncher(BinLauncher.kt:124)
        at com.jonnyzzz.mcpSteroid.devrig.BinLauncherKt.ensureBinLauncher(BinLauncher.kt:90)
        at com.jonnyzzz.mcpSteroid.devrig.MainKt$mainImpl2$2.invokeSuspend(Main.kt:93)
        ...
```

## Root cause
`ensureWindowsPathEntry` (`npx-kt/.../devrig/BinLauncher.kt:307-312`) spawns PowerShell to update `HKCU\Environment\Path`, but starts the process **incorrectly**:

```kotlin
ProcessBuilder("powershell", "-NoProfile", "-NonInteractive", "-Command", script)
    .redirectErrorStream(false)
    .redirectOutput(ProcessBuilder.Redirect.DISCARD)   // stdout discarded
    .redirectError(ProcessBuilder.Redirect.INHERIT)    // stderr inherited
    .start()
    .waitFor()                                         // <-- unbounded
```

- **stdin is never closed.** There is no `.redirectInput(...)`, so the child's stdin defaults to an open `Redirect.PIPE` that the parent never writes to or closes. PowerShell can block on that open stdin handle, so `waitFor()` never returns and the coroutine worker is pinned.
[...]

PR #282, opened 2026-07-16
installer: close stdin on 'devrig install devrig' (non-interactive contract)

jonnyzzz:
## What

Close stdin on the `devrig install devrig` handoff in both generated bootstrap installers, and document the non-interactive contract on the command it invokes.

## Why

When the installer is bootstrapped via `curl | sh` (POSIX) or `irm | iex` (PowerShell), the parent shell's stdin is a **pipe still open to the outer shell** that ran `curl`/`irm`. Handing that open stdin down to the `devrig install devrig` subprocess means any stdin read — or an interactive prompt — inside devrig would **block forever with no user recourse**. This is part 2 of #273 (the arch-detect crash is part 1).

## Changes

- **`install.sh`** — redirect the launcher invocation's stdin from `/dev/null`.
- **`install.ps1`** — pipe `$null` into the launcher invocation.
- **`runInstallDevrigCommand`** — document the contract: this path is invoked with an already-closed stdin, so **no stdin reads / prompts are permitted**; missing input must fail fast (non-zero), never block.
- **Unit tests** — pin both stdin closes on the actual launcher line.

## Contract

`devrig install devrig` is contractually **non-interactive**: no `readln`/`Console.readLine`/prompts on that path. Every input arrives as an explicit flag (`--install-script`, `--jdk-home`); anything missing fails fast rather than waiting for a user who cannot answer.

This is the foundational PR — the PowerShell arch/progress fixes (#273/#274, PR #281) build on top.

### DIFF
--- a/installer-gen/src/main/resources/templates/install.ps1.tmpl
+++ b/installer-gen/src/main/resources/templates/install.ps1.tmpl
@@ -115,7 +115,12 @@ if (-not (Test-Path (Join-Path $jdkHome 'bin\java.exe'))) {
 # registers itself. JAVA_HOME is set only so the unpacked devrig runs under the bundled JDK right now.
 Write-Log "registering devrig (devrig install devrig)..."
 $env:JAVA_HOME = $jdkHome
-& $launcher install devrig "--install-script=$launcher" "--jdk-home=$jdkHome"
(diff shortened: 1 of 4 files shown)


jonnyzzz:
It also fixes #150

PR #328, opened 2026-07-20
Issue 251 arena project name

### LINKED ISSUE #251: Arena MCP arm: hard-coded `project-home` name breaks the mandatory first steroid_execute_code call
## Summary

The DPAIA arena's MCP-arm prompt hard-codes the IntelliJ project name as the literal string
`project-home`, but the project actually open in the arena IDE has a **dynamic, per-run name**. This
makes the **mandatory first `steroid_execute_code` call fail** on every MCP-arm run, and the agent then
abandons MCP for the rest of the task. It is the root cause of the confound identified in the #250
review quorum — the "MCP arm" never meaningfully exercises the MCP transport.

Parent: #250

## Evidence (from the #250 gist)

`petclinic27-mcp-decoded.txt` — the agent's very first (prompt-mandated) `steroid_execute_code` call
returns:

```
ERROR: Project not found: "project-home". Available: demo-project-ki3y1pmc
```

The agent then explicitly concludes *"The IntelliJ project is a different demo project"* and does **all**
remaining work through Bash / Read / Write — MCP is called only **twice** in the whole run (both at the
start; the first fails), identical fallback behavior to the `none` arm. See #250 for the full turn-by-turn
breakdown and the three-reviewer quorum, all of which independently landed on this.

## Root cause

In `test-experiments/.../arena/ArenaTestRunner.kt`:

- **Line 37** — the project is cloned to a directory with a **dynamic per-run suffix**:
  ```kotlin
  val suffix = System.nanoTime().toString(36)
  val projectDir = "$projectGuestDir/${testCase.repoName}-$suffix"
  ```
  so the IntelliJ project name is dynamic (e.g. `spring-petclinic-<suffix>` / `demo-project-<suffix>`),
  never a fixed string.

- **Line 187** — the MCP-arm prompt nonetheless tells the agent the name is a **static literal**:
  ```kotlin
  appendLine("- **Project name in IntelliJ is always `project-home`** — use this exact name in every `steroid_execute_code` call. Never use the GitHub repo name (e.g. \"petclinic\", \"spring-petclinic\") as the project name.")
  ```

- **Line 105** separately (and correctly) gives the agent the project **path** (`$projectDir`), so the
  prompt is internally inconsistent: a correct interpolated path plus a wrong hard-coded name.

The observed `demo-project-ki3y1pmc` (note the `demo-project` prefix — cf.
`test-integration/src/test/docker/test-project/settings.gradle.kts:1` `rootProject.name = "demo-project"`)
suggests a **second, deeper problem to confirm**: the arena may not be opening the freshly-cloned task
project in the IDE at all for the MCP arm — the container's default `demo-project` is what's open, so
`steroid_execute_code` would be operating against the wrong project even if the name were correct.

## Impact

- Every MCP-arm run wastes its mandatory first call on a guaranteed error and then falls back to native
  tools → **the arena's MCP measurements are invalid** (they measure "MCP-prompt tax + bash", not the
  MCP transport). This directly undermines the cost comparison in #250.
- The prompt rule at line 187 is unreachable/misleading for agents and should never have shipped as a
  static constant given line 37.

## Proposed fix

1. **Interpolate the real project name into the prompt** instead of hard-coding `project-home`. Derive it
   from the same source as the open project (the dir basename `${testCase.repoName}-$suffix`, or better,
   query it) and inject it at line 187, e.g.:
   ```kotlin
   appendLine("- **Project name in IntelliJ is `$projectName`** — use this exact name in every `steroid_execute_code` call.")
   ```
   or instruct the agent to discover it once via `ProjectManager.getInstance().openProjects` /
   `project.name` (the prompt already prints `project.name` at line 239) and reuse it.
2. **Confirm and fix the project-open step** — verify the cloned task project (not the container default
   `demo-project`) is the one actually opened/indexed in the IDE before the agent starts. If it isn't,
   open it (or reuse the existing project window) as part of arena setup / `prewarm`.
3. **Fail fast in the harness** — if the mandated first `steroid_execute_code` returns
   `Project not found`, the arena should surface it as a setup error rather than letting the run silently
   degrade to the bash fallback and report a "successful" MCP arm.

## Acceptance criteria

- [ ] An MCP-arm run's mandatory first `steroid_execute_code` call succeeds (no `Project not found`).
- [ ] The project the agent operates on is the cloned+patched task project, verified by `project.basePath`
      matching `$projectDir`.
- [ ] A regression check (or an assertion in `ArenaTestRunner`) fails the run if the first steroid call
      cannot resolve the project, instead of degrading silently.
- [ ] Re-running `dpaia__spring__petclinic-27 (mcp)` shows MCP actually used for IDE work (build/PSI/test),
      not 2 calls then bash — a prerequisite for the de-confounded 2×2 experiment proposed in #250.

matveidetonator:
fixed #251 
now arena discovers project_name through steroid_list_projects, selects it by path, verifies that first execute_code targets right project. if project resolution or mcp execution fails, the arena fails the run instead of silently falling back to bash
a full arena run completed successfully, tests passed

### DIFF
--- a/test-experiments/src/test/kotlin/com/jonnyzzz/mcpSteroid/integration/arena/ArenaOutputParsing.kt
+++ b/test-experiments/src/test/kotlin/com/jonnyzzz/mcpSteroid/integration/arena/ArenaOutputParsing.kt
@@ -1,0 +1,118 @@
+/* Copyright 2025-2026 Eugene Petrenko (mcp@jonnyzzz.com); Copyright 2025-2026 JetBrains. Use of this source code is governed by the Apache 2.0 license. */
+package com.jonnyzzz.mcpSteroid.integration.arena
+
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonArray
+import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.JsonPrimitive
+import kotlinx.serialization.json.booleanOrNull
+import kotlinx.serialization.json.contentOrNull
+
+/** The plugin's project-resolution error prefix (ProjectScopedToolHandler.resolveProject). */
+private const val PROJECT_NOT_FOUND_PREFIX = "Project not found: \""
+
+/** Non-throwing string accessor: null unless this element is a JSON string/primitive. */
+private fun JsonElement?.stringOrNull(): String? = (this as? JsonPrimitive)?.contentOrNull
+
+/** Non-throwing boolean accessor: null unless this element is a JSON boolean primitive. */
+private fun JsonElement?.boolOrNull(): Boolean? = (this as? JsonPrimitive)?.booleanOrNull
+
+/**
+ * Extract text from a `content`/result field that may be EITHER a bare JSON string OR a JSON array of
+ * `{type,text}` content blocks (the valid raw Anthropic content-block shape). Non-throwing: safe casts
+ * only, so an unexpected shape yields "" rather than crashing the arena run.
+ */
+private fun textOf(el: JsonElement?): String =
+    (el as? JsonPrimitive)?.contentOrNull
+        ?: (el as? JsonArray)?.joinToString("\n") { (it as? JsonObject)?.get("text").stringOrNull() ?: "" }
+        ?: ""
+
+/**
+ * True when the agent's raw NDJSON contains a `steroid_execute_code` tool RESULT that failed with the
+ * plugin's project-resolution error. Structural on purpose:
+ *  - reads the raw NDJSON (never the output-filtered prose, which echoes the prompt and fetched articles);
+ *  - normalizes the wrapping `ERROR:` / `Error:` prefixes before an EXACT prefix match, so it matches the
+ *    real payload `ERROR: Project not found: "…"` (and the doubled `Error: ERROR: …` sibling field);
+ *  - attributes the error to `steroid_execute_code` via a tool-call-id -> tool-name map, because the
+ *    Claude/Gemini error result carries only `tool_use_id`/`tool_id`, no tool name. That mapping is what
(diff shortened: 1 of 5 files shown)


jonnyzzz (commented):
Overall the design is solid and headed in the right direction — enum-based status types, id-based attribution, and the general robustness are all good calls. A few API-shape refinements would be worth making before merge.

**1. Derived-property asymmetry.** `usedMcpSteroid` and `successfulMcpExecution` are `val` properties on `AgentTranscript`, but `projectResolutionStatus()` is a zero-arg, side-effect-free extension function. These are the same kind of derived value expressed in two different idioms. Making `projectResolutionStatus` a `val` property too would keep things consistent. (A parameterized helper like `firstExecutionTargetsProject(dir)` staying a function is perfectly fine, since it takes an argument.)

**2. Two parallel lists.** `executeCodeCallIds` + `executeCodeResults` force the caller to correlate entries by callId manually — which is exactly what `projectResolutionStatus` ends up doing. A single ordered list of calls, where each element carries its own optional result, would be more cohesive and remove that correlation burden from callers.

**3. `shouldFailRun` on the enum.** Putting `shouldFailRun` on the enum bakes runner policy into the type/status, which is a slight responsibility leak. Keeping the status pure and leaving the pass/fail decision to the caller is the more classic separation.

**4. Prompt parsing coupling.** `firstExecutionTargetsProject` parses `base:` out of free-form prompt text, which tightly couples it to the print format. Extracting the shared format into a constant (or at least documenting it) would ensure the producer and the parser can't silently drift apart.

**5. Redundant parameter.** In `evaluate(testCase, agentResult, agentResult.rawStdout, projectDir)`, `rawStdout` is already reachable via `agentResult`, so the third argument is redundant and can be dropped.

Nice work overall — once 1 and 5 (and ideally 2–3) are addressed, this is good to merge.

PR #356, opened 2026-07-24
feat(mcp-core): expose CLI metadata from tool specs

matveidetonator:
Rebased onto `main`.

Every `steroid_*` tool should also be a devrig subcommand generated from the tool's own metadata, instead of a hand-written Clikt command class per tool. Review ask on #272: "I can add a new MCP tool which will automatically (or at minimal cost) be registered to the CLI as well", and "the command parameters are easy to generate from the MCP Tool specification".

This PR is the `mcp-core` half of that: a typed CLI projection declared beside the MCP wire schema, plus the declarations filled in for every tool. **Declaration only** — no tool behavior changes, and no CLI library appears in either module. Clikt stays an implementation detail of the npx-kt frontend, which consumes these descriptors in follow-up work.

### What it adds

- A CLI projection on the tool spec, alongside (never replacing) the MCP wire schema. `devrigToolSpecs()` becomes the single tool-spec list, so later help, parsing, and runtime layers consume the same command descriptors without duplicating tool registration.
- Per parameter: a mandatory one-line `cliSynopsis`, so no help text is ever derived by clipping an MCP description; curated `cliMissingHint` wording; CLI-only `cliMinimum`/`cliMaximum` bounds, which exist because putting `minimum` in the schema's extra block would change the MCP wire; and `cliFileSource` for a flag that feeds file or stdin content into the parameter it belongs to.
- Per tool: `CliCommandSpec.extraOptions` for tool-scoped CLI options that are not tool inputs — currently `open_project`'s `--wait` polling toggle. `CliExtraOption` splits `name` (identity) from `flag` (presentation, defaulting to `--$name`), mirroring `InputSchemaParamSpec`, so downstream consumers key on the stable name instead of a literal flag string; construction-time validation matches the sibling `cliFileSource` builder.
- Declarations for all eight tools in `devrigToolSpecs()`.

### Why the three CLI-only flags have three different shapes

They are three unlike things. `--code-file` is an alternate source for `code`, so it is declared on `code` itself and exclusivity follows from having two sources for one parameter. `--wait` is per-tool orchestration, so it is a tool-level extra. `--out` is a render-path redirect applied after the tool has already returned, so it belongs to the CLI frontend and is declared nowhere here. The first two need no per-tool code in the generator that consumes this metadata.

One surface rule covers every CLI-only declaration: never gate it per surface, because it never reaches the MCP wire. Only wire-visible parameters, such as `open_project`'s `backend_name`, stay gated by `includeBackendName`.

### The MCP wire is unchanged

A golden test pins every tool's `inputSchema` byte-for-byte, with description values masked by a JSON-tree walk — so CLI metadata cannot leak into the wire, while routine prompt-copy edits do not churn the test. The test documents how to regenerate the constants legitimately.

`TODO.md` records the two behaviors the CLI frontend must now own.

### Size

18 files, +1301/−18 — of which production code is +400/−18 across 12 files (mostly one declarative block per tool); tests are +892 across 5 files, dominated by golden constants.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -98,6 +98,15 @@
   IDE failing its `/windows` fetch errors the whole call (`coroutineScope` + `error(...)`), unlike
   `list_projects` which degrades per-backend. Return partial windows + a per-backend error marker.
 
+- [ ] **devrig CLI must own `--out` and the `--wait` polling loop (#284)**: the schema-driven-command
+  reshape removed the `out` parameter from `VisionScreenshotToolSpec` and turned `--wait` into a
+  declared `CliExtraOption` on `steroid_open_project`, because neither is a tool input. The tool
+  metadata therefore no longer carries either behavior: the CLI frontend must document `--out` as a
+  framework flag beside `--json` (a render-path redirect writing a returned image to a path — note
+  `steroid_execute_code` also returns PNGs via `logImage`, e.g. the modal-dialog failure screenshot,
+  so it is not screenshot-only) and must implement `--wait` as a `list_windows` poll until the project
+  reports initialized. Without those two, both flags silently vanish from the CLI.
+
 - [ ] **red-code reporter false-positives on Kotlin files**: `reportProjectRedCode` (PSI reference scan,
   `mcp-steroid-import.kt`) reports Kotlin stdlib/operator references (`mutableMapOf`, `runCatching`,
   `trim()`, `!!`, `=`) as UNRESOLVED — 95/646 on the stock Gradle test-project's `SsrRunCatchingDemo.kt`
--- a/TODO.md
+++ b/TODO.md
@@ -107,6 +107,19 @@
   so it is not screenshot-only) and must implement `--wait` as a `list_windows` poll until the project
   reports initialized. Without those two, both flags silently vanish from the CLI.
 
+- [ ] **Harden the CLI tool-spec metadata layer (#284 follow-up)**: three review findings deferred from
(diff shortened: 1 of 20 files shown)


jonnyzzz (approved):

jonnyzzz:
Thank you for this contribution — merging. The design is exactly what #272/#284 asked for: a declaration-only CLI projection beside the MCP wire schema, no CLI library anywhere near `mcp-core`, and it held up cleanly under a three-lens review — the wire is provably unchanged, and all eight golden constants were independently re-derived and then re-captured from main (`75c611b1`) byte-for-byte.

Four small commits were applied on top before merging:

- `DevrigToolSpecsGoldenSchemaTest`'s KDoc cited a capture commit that isn't an ancestor of `main` and a description-pinning backstop that doesn't exist (the per-tool tests only assert non-blankness) — rewritten to the verified provenance and the honest masking trade-off.
- `StubStdioMcpServer` now registers from `devrigToolSpecs()`, so the canonical list has a production consumer and the served MCP surface can't drift from what the generated CLI will consume (`DevrigDescriptorParityTest` mirrors the new callsite).
- `FetchResourceToolHandler`'s logger went back to eager — the lazy wrapper wasn't needed.
- Deferred hardening notes landed in `TODO.md`: a read-only schema view (mutable `ToolSchema` is reachable via `CliToolSpec.schema`), declaration-time flag-collision validation, and wire-bound visibility in `asCliParams()`.

Looking forward to the npx-kt generator half!

Issue #10, opened 2026-02-09
Kotlin compiler version mismatch: scripts fail to compile against IntelliJ with Kotlin 2.4 plugin


jonnyzzz:
## Summary

The MCP-Steroid plugin's embedded Kotlin compiler (metadata version **2.2.0**) is incompatible with the Kotlin plugin libraries shipped in recent IntelliJ IDEA builds (metadata version **2.4.0**). This causes **every** `steroid_execute_code` invocation that goes through the `kotlinc` compilation path to fail — including trivially simple scripts like `println("test")`.

## Error

Every compilation attempt produces hundreds of lines like:

```
module was compiled with an incompatible version of Kotlin.
The binary version of its metadata is 2.4.0, expected version is 2.2.0.
```

The errors originate from `.kotlin_module` files inside `/Applications/idea.app/Contents/plugins/Kotlin/lib/` which are on the compilation classpath.

## Reproduction

1. Install IntelliJ IDEA with the bundled Kotlin plugin (metadata version 2.4.0)
2. Install MCP-Steroid plugin
3. Execute any script via `steroid_execute_code`, even:
   ```kotlin
   println("test")
   ```
4. Compilation fails with the metadata version mismatch error

## Root Cause

The `kotlinc` bundled with (or used by) MCP-Steroid expects metadata version **2.2.0**, but the IntelliJ Kotlin plugin JARs on the classpath were compiled with Kotlin **2.4.0**. Since the wrapper code generated by MCP-Steroid imports IntelliJ platform classes (`com.intellij.openapi.project.*`, `com.intellij.openapi.application.*`, `com.jonnyzzz.mcpSteroid.execution.McpScriptContext`), the compiler must read metadata from those JARs and fails.

Kotlin's forward-compatibility window is **+1 minor version only** (e.g., 2.2 can read 2.3, but **not** 2.4). The 2.2 → 2.4 gap is outside the supported range.

## Why `-Xskip-metadata-version-check` is not a viable fix

- The flag only suppresses the error, it doesn't make the compiler understand newer metadata formats
- For a **+2 version gap**, behavior becomes undefined and may cause strange compilation or runtime errors
- `kotlin-reflect` relies on metadata at runtime; mismatches can produce incorrect behavior
- Even Gradle has [planned to remove their usage of this flag](https://github.com/gradle/gradle/issues/26810)

## Suggested Fix

[...]

PR #359, opened 2026-07-27
fix: bump bundled kotlinc 2.3.20 → 2.4.10 for IDE 263 compatibility

### LINKED ISSUE #357: Regression of #10: bundled kotlinc 2.3.20 cannot compile against IntelliJ IDEA 2026.3 EAP (Kotlin metadata 2.5.0) — every execute_code fails
### Summary

`steroid_execute_code` fails for **every** script, including a one-liner, on IntelliJ IDEA 2026.3 EAP (`IU-263.2180`). The bundled kotlinc is 2.3.20 while the IDE's bundled Kotlin plugin jars carry metadata **2.5.0**, so the compiler cannot read the IntelliJ platform classpath.

This is the same failure mode as #10 (closed 2026-02-13, 2.2 vs 2.4), reopened by a new gap. The fix for #10 was a manual pin bump (48c8131a, "bump Kotlin 2.2.20→2.3.20 + ktor 3.1.0→3.3.2 (match IDE 261 bundled)"), not the version-detection approach #10 itself proposed, so the same break recurs on each IDE Kotlin bump.

### Environment

| | |
|---|---|
| Plugin | `0.101.566-jb-90ac87b` |
| IDE | IntelliJ IDEA 2026.3 EAP, `IU-263.2180` |
| IDE Kotlin plugin | `263.2180.0-IJ`, class metadata **2.5.0** |
| Bundled kotlinc | **2.3.20** (`kotlin-cli/build.gradle.kts:27`, origin/main @ 40d32ebc) |
| OS | Windows 10 Pro 19045 |
| Unaffected on same machine | IntelliJ IDEA 2026.1.4 (`IU-261.26222.65`), 2026.2 (`IU-262.8665.337`) |

Read from `~/.mcp-steroid/markers/23684.mcp-steroid`:

```json
"ide":    { "name": "IntelliJ IDEA 2026.3 EAP", "version": "2026.3 EAP", "build": "IU-263.2180" },
"plugin": { "id": "com.jonnyzzz.mcp-steroid", "name": "MCP Steroid", "version": "0.101.566-jb-90ac87b" }
```

### Reproduction

Against the 2026.3 EAP instance, any script:

```kotlin
"OK build=" + com.intellij.openapi.application.ApplicationInfo.getInstance().build.asString()
```

fails before running, with hundreds of lines of:

```
execution_id: eid_20260727T072353-bridge-probe
Compiler Errors/Warnings:
C:\Users\serge\AppData\Local\Programs\IntelliJ IDEA Ultimate 3\plugins\Kotlin\lib\kotlin-plugin.jar!\META-INF\kotlin-script-runtime.kotlin_module:
  error: module was compiled with an incompatible version of Kotlin.
  The binary version of its metadata is 2.5.0, expected version is 2.3.0.
C:\...\kotlin-plugin.jar!\META-INF\org.jetbrains.kotlin_analysis-api-fir.kotlin_module: ... 2.5.0, expected 2.3.0
C:\...\kotlin-plugin.jar!\META-INF\org.jetbrains.kotlin_checkers.kotlin_module:        ... 2.5.0, expected 2.3.0
C:\...\kotlinc.scripting-compiler-plugin.jar!\META-INF\org.jetbrains.kotlin_kotlin-scripting-compiler.kotlin_module: ... 2.5.0, expected 2.3.0
(… ~200 more, every module in plugins/Kotlin/lib/)
```

The same script succeeds immediately against the 2026.1 and 2026.2 instances on this machine, so it is build-specific and not environmental.

This instance worked on **2026-07-22** (compiled artifacts still present under `%LOCALAPPDATA%\JetBrains\IntelliJIdea2026.3\tmp\mcp-steroid\mcp-steroid\eid_20260722T*/compiled/script.jar`) and broke after the EAP moved to 263.2180.

### Root cause

`kotlin-cli/build.gradle.kts:27` hardcodes the compiler that ships in the plugin:

```kotlin
val kotlincVersion = "2.3.20"
```

Kotlin's metadata read window is +1 minor. 2.3 reading 2.5 is two minors ahead and rejected. This matches the repo's own rule in `buildSrc/.../KotlinVersionCompatibility.kt`:

```kotlin
fun isCompatible(ideKotlin: KotlinVersion, bundledKotlin: KotlinVersion): Boolean {
    if (ideKotlin.major != bundledKotlin.major) return false
    return abs(ideKotlin.minor - bundledKotlin.minor) <= 1
}
```

`2.5` vs `2.3` gives `abs(5 - 3) = 2 > 1`, so the plugin already classifies this pairing as unsupported. The gap is that `VerifyBundledKotlinCompatibilityTask` runs at **build** time against the IDE the plugin is built against. It cannot see the IDE a user actually attaches to, so a user on a newer EAP gets a wall of compiler errors instead of that verdict.

### Impact

Total loss of `steroid_execute_code` on the affected IDE, which takes down everything built on it. In our case that is JetDesk's IDE bridge (`scripts/lib/ide-bridge.js`), whose default `prefer: 'monorepo'` scoring picks exactly this instance because it has the IntelliJ monorepo open. All snapshot, thread-dump and settings-path tooling routed through it fails until the instance is closed or a port is pinned manually.

The failure is also hard to attribute from the client side. The first ~200 characters of the error are truncated in most MCP clients, so what surfaces is a compiler error about `kotlin-script-runtime` rather than "this IDE build is unsupported".

### Fix direction

1. **Bump the pin** to a kotlinc matching 2026.3's Kotlin (2.5.x) as the immediate unblock. This keeps working until the next IDE bump, which is the pattern #10 already established.
2. **Detect and fail fast.** Evaluate `KotlinVersionCompatibility.isCompatible` at **runtime** against the attached IDE's Kotlin version, once at plugin startup or on the first `execute_code`. When it fails, return a single actionable line instead of the compiler dump, for example: `mcp-steroid 0.101.566 bundles kotlinc 2.3.20, this IDE (IU-263.2180) bundles Kotlin 2.5.0. Unsupported combination (max 1 minor apart). Update the plugin or use an IDE 2026.1/2026.2.` This is what #10's "detect the Kotlin version from the running IDE's classpath" suggestion asked for and is the part that would stop this from recurring silently.
3. **Ship more than one compiler, or resolve it at runtime,** so a single plugin build spans the release plus the current EAP. Otherwise every EAP Kotlin bump breaks every user on that EAP until a plugin release lands.
4. **Surface it in the tool listing.** Since `steroid_list_projects` / `steroid_list_windows` still answer normally while `execute_code` is dead, a client cannot tell a healthy instance from a broken one without compiling. A health flag on the list responses would let clients route around a broken instance. Related to #155, which asks for IDE identity on those same responses.

### Related

- #10 — same failure at 2.2 vs 2.4, closed as completed via the manual pin bump
- #155 — host-IDE identity missing from list responses, same clients-cannot-detect-instance-health theme

CrazyCoder:
Fixes #357.

## What

One line: `kotlin-cli/build.gradle.kts` `kotlincVersion` `2.3.20` → `2.4.10`, plus a comment recording the constraint so the next bump has the rule in front of it.

## Why

IntelliJ IDEA 2026.3 EAP bundles Kotlin **2.5.0**. Kotlin reads class metadata at most **+1 minor** ahead, so the bundled 2.3.20 compiler cannot read a single module under `plugins/Kotlin/lib`. Every `steroid_execute_code` call fails before the script body runs, including a one-liner:

```
kotlin-plugin.jar!/META-INF/kotlin-script-runtime.kotlin_module:
error: module was compiled with an incompatible version of Kotlin.
The binary version of its metadata is 2.5.0, expected version is 2.3.0.
```

…repeated for ~200 modules. Observed on `IU-263.2180` and `IU-263.2289` with plugin `0.101.566-jb-90ac87b`. The same scripts run fine against 2026.1 (`IU-261.26222.65`) and 2026.2 (`IU-262.8665.337`) on the same machine.

This is the same failure #10 hit at 2.2 vs 2.4.

## Why 2.4.10

It is the newest publicly released kotlinc — there is no 2.5.x on GitHub releases or Maven Central yet; 2026.3 EAP ships a pre-release 2.5.0. 2.4.10 sits within one minor of **both** ends of the supported range, so a single bundled compiler spans every IDE the plugin targets:

| IDE line | Bundled Kotlin | 2.4.10 reads it |
|---|---|---|
| 261 / 262 | 2.3.20 | backwards, always fine |
| 263 EAP | 2.5.0 | one minor forward, inside the window |

`verifyBundledKotlinCompatibility` passes and independently agrees:

```
IDE Kotlin version: 2.3.20
Kotlin plugin version: 2.3.20
Bundled kotlinc version: 2.4.10
Kotlin plugin not newer than IDE: true
Bundled kotlinc compatible (major match, minor diff <= 1): true
Latest stable Kotlin release: 2.4.10 (bundled kotlinc is up to date)
```

The root `kotlin("jvm") version "2.3.20"` is deliberately untouched — that compiles the plugin itself, and older metadata is always readable by a newer runtime.

## Verification

Built and installed locally into a 2026.3 EAP (`IU-263.2289`), replacing `0.101.566-jb-90ac87b`. Before the change every script failed as above. After:

```kotlin
val app = com.intellij.openapi.application.ApplicationInfo.getInstance()
println("IDE=" + app.fullApplicationName + " build=" + app.build.asString())
```
```
IDE=IntelliJ IDEA 2026.3 EAP build=IU-263.2289
```

Full `steroid_execute_code` and the javaperf/Profiler script paths work again.

## On the missing test

CONTRIBUTING asks for a failing test with a bug fix, and I did not add one — flagging that rather than bolting on something that passes trivially.

The change is a build constant, and the check that *would* own this regression already exists: `VerifyBundledKotlinCompatibilityTask`. It cannot catch this case because it compares the bundled kotlinc against the Kotlin of the IDE the plugin **builds** against (261 → 2.3.20), not the IDE a user **attaches** to. That is why the build stayed green while every 263 user was broken, and it is why a pin bump alone will break again on the next IDE Kotlin bump.

#357 lays out the durable fixes — evaluate `KotlinVersionCompatibility.isCompatible` at runtime against the attached IDE and fail with one actionable line instead of the compiler dump, and/or resolve the compiler per-IDE. Both are larger than this PR and mix a fix with build-infra or shipped-runtime changes, so I kept this one to the unblock per the "one logical change per PR" guideline. Happy to follow up with either if you tell me which direction you prefer.

### DIFF
--- a/kotlin-cli/build.gradle.kts
+++ b/kotlin-cli/build.gradle.kts
@@ -24,7 +24,11 @@ tasks.test {
 
 // --- kotlinc download and distribution ---
 
-val kotlincVersion = "2.3.20"
+// Must stay within 1 minor of the Kotlin bundled by the IDEs we attach to
+// (Kotlin reads metadata at most +1 minor ahead). IDE 263 bundles Kotlin
+// 2.5.0, so 2.3.x is 2 minors behind and fails to read plugins/Kotlin/lib.
+// 2.4.10 is the newest publicly released kotlinc and covers 2.3-2.5 IDEs.
+val kotlincVersion = "2.4.10"
 val kotlincUrl = "https://github.com/JetBrains/kotlin/releases/download/v${kotlincVersion}/kotlin-compiler-${kotlincVersion}.zip"
 val kotlincSha256Url = "$kotlincUrl.sha256"
 val kotlincDownloadDir = layout.buildDirectory.dir("kotlinc-zip/$kotlincVersion")


jonnyzzz:
Verified the pin choice against the compiled bytecode of the latest internal builds of all three IDE lines (`IU-261.27181`, `IU-262.9726`, `IU-263.2487` — current master head, newer than the builds in the report). Full measurement table and method in https://github.com/jonnyzzz/mcp-steroid/issues/357#issuecomment-5130667855. Summary of what matters for this pin:

| IDE line | `plugins/Kotlin/lib` metadata (measured) | kotlinc 2.3.20 reads it | kotlinc **2.4.10** reads it |
|---|---|---|---|
| 261 | 2.4.0 | yes (+1) | yes (backwards-plus) |
| 262 | 2.4.0 | yes (+1) | yes |
| 263 | **2.5.0** | **no (+2)** | yes (+1) |

Two confirmations beyond the PR text:

- The 2.5.0 metadata on 263 persists on current master (`263.2487`), and it lives **only** in `plugins/Kotlin/lib` — the 263 platform jars themselves are metadata 2.4.0. So the constraint driver is exactly the Kotlin plugin artifacts on the script classpath, and 2.4.10 is currently the only single pin that spans all three lines.
- 2.4.10 is confirmed the newest stable on JetBrains/kotlin GitHub releases (v2.4.10, 2026-07-14; v2.4.20-Beta2 is pre-release), i.e. this is 2.4.<latest stable> as we want it.

Agreed on keeping the runtime `isCompatible` check + actionable one-line error as a follow-up rather than in this PR — that is the durable half of #357 (a 2.4.x pin stops reading at 2.5, so the same break returns when the 263/264 Kotlin plugin moves to 2.6-dev metadata).

Running the compatibility compile tests against 261 / 262 / 263 with this patch applied next; will report results here.

jonnyzzz:
Test results with this PR applied — the KtBlock compilation matrix (`:prompts:test`): all 355 prompt-article Kotlin fences compiled by the **bundled kotlinc against the full IDE classpath**, i.e. the same compile path `steroid_execute_code` takes at runtime. Run against the latest internal builds of all three lines:

| IDE build | Mode | Result |
|---|---|---|
| `IU-261.27181` | strict (`-Werror`) | **355/355 pass** |
| `IU-262.9726` | EAP (warnings allowed) | **355/355 pass** |
| `IU-263.2487` | EAP (warnings allowed) | **355/355 pass** |

Each run ~7 min, fresh compilations (build cache disabled), kotlinc 2.4.10 downloaded from GitHub releases with sha256 verification as part of the run. `:kotlin-cli:test` also green.

A/B control on the **identical full 263 classpath** to confirm the matrix is sensitive to this exact failure:

- kotlinc **2.3.20**: 135 modules fail with `module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 2.5.0, expected version is 2.3.0` — first of them `kotlin-plugin.jar!/META-INF/kotlin-script-runtime.kotlin_module`, byte-for-byte the failure in #357.
- kotlinc **2.4.10**: clean compile, exit 0.

Merge preview onto current `main` is conflict-free (`main`'s recent `downloadKotlinc`/`kotlincDistElements` refactor touches different hunks). Landing after the plugin-build gate completes. Thanks for the precise report and the fix, Serge — textbook.

jonnyzzz:

PR #361, opened 2026-07-27
Switch Kotlin snippets compilation to Kotlin Build Tools API

### LINKED ISSUE #207: Script compilation is 94% of execute_code wall time (avg 7.8s per call) — speed it up
Data from the 2026-06-26 full eval run (141 tasks, plugin `0.101-40690055`, 541 `execute_code` executions).

## Measurements

Timing from per-execution artifact mtimes (`script.kts` write → `compilation-success.txt` → `output.jsonl`; 1-second granularity, 515 executions with a compile marker):

| Phase | avg | median | p90 | min | max |
|---|--:|--:|--:|--:|--:|
| Compile | **7.8s** | 7.0s | 11.0s | 5s | 23s |
| Script run | 0.5s | 0.0s | 1.0s | — | — |

- **Compilation is 94% of the measured execution window** (request persist → output written). The scripts themselves are near-instant.
- **Total compile time across one run: ~67 aggregate compile-minutes** (tasks ran in parallel, so this is summed compute wait, not elapsed run time).
- The **5s floor** with a ~10-line script says the cost is fixed overhead, not code size.
- Distribution: 290 executions at 5–8s, 218 at 8–15s, 7 at ≥15s.

Caveat: the measured window also contains the pre-compile `awaitRefresh` (VFS sync), so "compile" here is "everything between request persist and compile-success". The 5s floor on trivial scripts still points at process startup as the dominant term.

## Why it is slow (verified in source)

- Every call spawns a **fresh `kotlinc` process** — cold JVM + compiler init each time (`KotlincProcessClient.kt`, `GeneralCommandLine(kotlinc)`).
- The classpath is the **whole IDE: 1,621 jars** (~124 KB `-classpath` string in `kotlinc.args`).
- Compilations are **serialized** behind `compilationMutex` (`CodeEvalManager.kt:40`) — concurrent `execute_code` calls queue behind each other.

## What would NOT help much

A compile cache keyed on script hash: only **6 of 541 scripts** in the run were exact duplicates. Agents write a new script every call.

## Directions worth evaluating

1. **Keep the compiler warm.** Kotlin compiler daemon, or a persistent kotlinc worker process, or the embeddable compiler in-process. This attacks the 5s floor directly.
2. **Shrink the classpath.** 1,621 jars is everything; scripts use a small, stable subset. Even a curated subset for the common case (with full-classpath fallback on unresolved reference) could cut compiler init and resolution time.
3. **Overlap instead of serialize.** The mutex is there for a reason, but pre-warming the next compile context while a script runs may cut wall-clock, if the mutex constraints can be preserved.

Why it matters: at ~4 steroid calls per task, compilation adds ~30s of dead time per task. The eval data shows agents already prefer built-in tools 10:1 — every second of `execute_code` latency is a reason to stay on Read/Edit. This is a direct usability lever.

## Evidence

Private eval-logs repo: `2026-06-29/` per-execution artifacts (`.idea/mcp-steroid/eid_*/`: `script.kts`, `compilation-success.txt`, `kotlinc.args`, `classpath.txt`, `output.jsonl`). Measurement script in the run REVIEW. Related: #138 (adoption depth — latency is part of why agents stay on built-ins).

Tapchicoma:
This PR drops the old download-and-invoke-kotlinc integration and replaces it with the Kotlin Build Tools API (BTA).

*(Description updated 2026-08-04 to match the reviewed/hardened state; the original draft embedded the impl jars as resources with runtime extraction and pinned impl 2.3.20 — see the review threads and perf comments for the full evolution.)*

## Architecture (final)

- **One implementation flow**: in-process compilation with BTA's BuildSession-owned compiler application-environment pin. No execution-policy toggle and no daemon flow (the daemon still clears its jar caches around every operation — KT-88183).
- **One field-owned BuildSession per project service**: it is closed at project shutdown. A clean-cache A/B across 18 real IDEA/PyCharm/CLion prompt compilations took **15.023 s shared vs 24.587 s per-compilation**; closing every session releases KT-87743's application-environment/JarFS pin and makes the next compilation rebuild those caches.
- **No compiler-specific timeout/cancellation machinery**: BTA's blocking operation runs on `Dispatchers.IO` and finishes normally. We no longer call `BuildOperation.cancel()`, translate `OperationCancelledException`, or special-case compiler timeout results.
- `kotlin-build-tools-api` is a plain dependency of `kotlin-cli` (plugin `lib/`, IDE classloader sees the API only). The implementation ships as real jars in the plugin distribution's `kotlinc/` folder. Nothing is embedded as resources or extracted at runtime; `KotlinBuildsSession` loads the jar paths into an isolated classloader.
- **Versions**: api == impl == **2.4.20-RC-197**, the numbered RC build containing KT-87743. BTA versions are direct literals where needed (no version catalog, Gradle property, or provider); the Kotlin Gradle plugin used by the rest of the project remains **2.3.20**.
- **Upstream application-environment pin (KT-87743)**: BTA owns the pin for each in-process BuildSession, so the local reflective workaround is deleted. Structural tests verify a stable environment/ref-count across compiles, flat jar-handler/heap behavior, and full release on session close.
- Registry compiler params are tokenized and fail fast; compiler messages are remapped to user-code coordinates with the offending source line, and raw diagnostics are persisted per execution.

## Performance

| | before | after |
|---|---|---|
| production exec_code warm compile | 3–6 s (kotlinc subprocess) | **~218 ms** (14–28×) |
| 356-snippet IDEA-stable KtBlock matrix, clean cache | 401 s | **54 s** (7.43× vs subprocess; 2.37× vs unpinned BTA) |

BuildSession lifetime A/B (18 real-classpath compilations, clean cache, same machine and code):

| lifetime | suite time |
|---|---:|
| shared session, closed after the final compilation | **15.023 s** |
| fresh session created and closed for every compilation | 24.587 s |

Fresh sessions took 1.64× as long; sharing the session cuts this suite time by 39%.

## Validation

- `:kotlin-cli:test`: **60/60** (including application-environment/JarFS lifecycle, 60-compile heap/cache flatness, repeated fresh-session cleanup, error storm, large input, spaces in paths, and >40K classpath).
- Focused real-classpath prompt suite: **18/18**; restored-cache verification completed in **6.921 s**.
- `:ij-plugin:compileKotlin`: passed after removing the timeout result path.
- IntelliJ inspections: clean on every edited Kotlin file.
- Previous full PR validation remains green: `:ij-plugin:test` 268 tests, plugin ZIP exact-set/class-file gates, and the full cross-IDE KtBlock matrix 1,940/1,940.

Durable performance notes and the remaining release/issue watch list are in `docs/bta-snippet-compilation-performance.md` (KT-57097, KT-88182, KT-88183, KT-88278).

Closes #207


ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/execution/CodeEvalManager.kt
```diff
@@ -34,10 +40,21 @@ inline val Project.codeEvalManager: CodeEvalManager get() = service()
...
+        kotlinBuildsSession.close()
+        @OptIn(ExperimentalPathApi::class)
+        try {
+            btaWorkingDir.deleteRecursively()
+        } catch (_: Exception) {}
+    }
 
     private val log = thisLogger()
     private val compilationMutex = Mutex()
+    private val btaWorkingDir = Files.createTempDirectory("kotlin-bta")
```
jonnyzzz: use the StorageManager to compute the temp path

ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/execution/CodeEvalManager.kt
```diff
@@ -77,60 +94,69 @@ class CodeEvalManager(
...
+                resultBuilder.reportFailed("kotlinc stopped on timeout")
+                return null
+            }
 
-                if (compilerError.isNotEmpty()) {
-                    // Log stderr output (warnings/errors) for transparency.
-                    // Actual failure is determined by the exit code check below,
-                    // since stderr may contain mere warnings that don't block compilation.
-                    resultBuilder.logMessage("Compiler Errors/Warnings:\n$compilerError")
+            val compilerMessages = compilerMessageRenderer.getRecordedMessages()
```
jonnyzzz, line 144: Let's persist compiler messages under the exec folder, and potentially copy them to the logger too

ij-plugin/src/test/kotlin/com/jonnyzzz/mcpSteroid/koltinc/KotlincCommandLineBuilderIntegrationTest.kt
```diff
@@ -1,257 +0,0 @@
...
-
-        kotlincProcessClient.kotlinc(commandLine.args)
-
-        assertTrue("Expected output jar at ${commandLine.outputJar}", Files.exists(commandLine.outputJar))
-        ZipFile(commandLine.outputJar.toFile()).use { zip ->
-            assertNotNull("Expected MainKt.class in jar", zip.getEntry("MainKt.class"))
-        }
-    }
-
-    fun testCompilesJarWithClasspathArgFileContainingSpaces(): Unit = timeoutRunBlocking(90.seconds) {
```
jonnyzzz, line 154: this was useful it seems, but it should rather call our component, there are tests for various edge cases -- whitespaces, windows, etc.

ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/execution/CodeEvalManager.kt
Tapchicoma: I could not find `StorageManager`, but I updated the code to use `project.getProjectCachePath()` method instead. Is it fine?

kotlin-cli/src/main/kotlin/com/jonnyzzz/mcpSteroid/koltinc/KotlinBuildsSession.kt
```diff
@@ -0,0 +1,189 @@
...
+                    operation = jvmOperation,
+                    executionPolicy = when (executionPolicy) {
+                        CompilationExecutionPolicy.IN_PROCESS -> inProcessExecutionStrategy
+                        CompilationExecutionPolicy.DAEMON -> daemonExecutionPolicy
+                    },
+                    logger = kotlinLogger,
+                )
+            }
+        } catch (e: CancellationException) {
+            jvmOperation.cancel()
```
ALikhachev: This may throw `org.jetbrains.kotlin.buildtools.api.OperationCancelledException` which is not handled anywhere now

kotlin-cli/src/main/kotlin/com/jonnyzzz/mcpSteroid/koltinc/KotlinBuildsSession.kt
```diff
@@ -0,0 +1,189 @@
...
+     * @param argumentsConf A lambda that allows configuration of additional JVM compiler arguments. '-no-stdlib' and '-no-reflect' arguments are always added.
+     * @return A [CompilationResult] encapsulating the result of the compilation process.
+     *
+     * @throws kotlinx.coroutines.TimeoutCancellationException on reaching configured timeout during compilation
+     */
+    suspend fun compileKotlin(
+        sources: List<Path>,
+        destinationDir: Path,
+        executionPolicy: CompilationExecutionPolicy = CompilationExecutionPolicy.DAEMON,
+        compilationTimeout: Duration = 120_000L.toDuration(DurationUnit.MILLISECONDS),
```
ALikhachev: Nit: why not just `120L.toDuration(DurationUnit.SECONDS)`?

kotlin-cli/src/main/kotlin/com/jonnyzzz/mcpSteroid/koltinc/KotlinBuildsSession.kt
```diff
@@ -0,0 +1,189 @@
...
+ */
+@OptIn(ExperimentalBuildToolsApi::class)
+class KotlinBuildsSession(
+    val workingDir: Path,
+    val kotlinLogger: KotlinLogger? = null,
+) : AutoCloseable {
+
+    private val buildToolsApi = KotlinToolchains.loadImplementation(getBtaImplJars())
+    private val inProcessExecutionStrategy = buildToolsApi.createInProcessExecutionPolicy()
+    private val daemonExecutionPolicy = buildToolsApi.daemonExecutionPolicyBuilder().build()
```
ALikhachev: What is the lifetime of the process?
If it dies often, maybe worth configuring the daemon in the detached mode: https://kotlinlang.org/docs/kotlin-daemon.html#lifetime

ALikhachev (changes requested):

jonnyzzz:
I'm going to rebase this branch to the fresh `main`, please be careful

jonnyzzz:
## Performance: measured and profiled

**A/B on the prompts KtBlock matrix** — 355 snippets compiled against the IDEA-stable classpath, clean compile cache, same machine, 8 parallel test forks:

| | main (kotlinc subprocess per snippet) | this PR (BTA, shared session) |
|---|---|---|
| wall clock | 401 s | **128 s → 3.1×** |
| summed per-test time | 1186 s | 523 s → 2.3× |

**Per-compile latency** (probe against the real IDE classpath — 1789 jars, wrapper + `McpScriptContext` sources, i.e. the exact `steroid_execute_code` workload):

| | cold (first compile of a session) | warm steady state |
|---|---|---|
| BTA in-process | ~2.1 s | **370–500 ms** |
| BTA **daemon** (production exec_code path) | ~2.3 s incl. daemon spawn | **350–580 ms** |
| control: same snippet, stdlib-only classpath | — | ~40 ms |
| old kotlinc subprocess (#207 pain) | — | 3–6 s |

So the production per-call path went from 3–6 s to ~0.4 s warm — **~8–13× per compile**. `loadImplementation` costs 30 ms; daemon RMI overhead vs in-process is not measurable.

**Where the remaining ~400 ms goes** (JFR `jdk.ExecutionSample` over the probe run):

| bucket | share |
|---|---|
| FIR resolve + analysis (resolving against the 1789-jar classpath) | 28% |
| jar-FS central-directory parse/index (`cli.jvm.compiler.jarfs.FastJar*`) | 23% |
| backend (Fir2Ir + JVM codegen) | 18% |
| classpath roots / package parts / module metadata | 8% |
| misc (JRT-FS, protobuf metadata, strings, classloading) | ~23% |

The stdlib-only control says it plainly: the same snippet compiles in 40 ms on a small classpath — **the dominant per-compile cost is classpath processing, not compilation**. The ~2 s cold-session cost is the same thing in bulk: building the compiler application environment and parsing the central directories of all 1789 jars, paid once per `BuildSession` (that's also why the test matrix pays it 8× — once per fork).

Why the matrix shows 3.1× rather than the per-call 8–13×: of the 128 s, pure compile time is only ~18 s; the rest is Gradle/codegen fixed costs, 8× cold sessions, and per-test infrastructure (classpath walk, cache-key hashing, temp IO). And on repeated runs the compile cache short-circuits unchanged snippets entirely.

What would push end-to-end past 10×: classpath-structure reuse across sessions/compilations inside BTA — @Tapchicoma this is precisely the degenerate case you described collecting use-cases for (one small source, huge constant classpath, `.gradle.kts`-shaped); a persistent jar/classpath index shared across `BuildSession`s would attack both the 2 s cold start and the ~30% classpath share of every warm compile.

jonnyzzz:
The JFR recording behind the profile numbers above: https://gist.github.com/jonnyzzz/ff9dc1b10f325eda9c7f990af9d3dcd7

- `bta-inprocess-probe.jfr` (5 MB, `settings=profile`) — the whole probe run: session init → 1 cold + 7 warm compiles against the 1789-jar IDEA-stable classpath → 3 stdlib-only control compiles → daemon-client phase → 2 compiles on a fresh session in the warm JVM.
- Direct download: https://gist.githubusercontent.com/jonnyzzz/ff9dc1b10f325eda9c7f990af9d3dcd7/raw/bta-inprocess-probe.jfr
- Open with JDK Mission Control, or: `jfr view hot-methods bta-inprocess-probe.jfr` / `jfr print --events jdk.ExecutionSample bta-inprocess-probe.jfr`

Top of `jfr view hot-methods` for the impatient: `jarfs.ZipImplementationKt.parseCentralDirectory` 4.3%, `HashMap.getNode` 4.0%, `jarfs.ByteArrayCharSequence.toString` 2.0%, `jarfs.FastJarHandler.createFile/splitPath/<init>` ~4.3% combined — the jar-FS layer is all over the profile, actual codegen frames (`Fir2IrVisitor`) show up at 0.7%.

jonnyzzz:
## Warm snippet compiles at ~220 ms: application-environment pin + in-process execution

Follow-up to the perf comments above — this lands the fastest safe configuration we measured, with the memory story verified in depth. For review by the team; all claims below are source/bytecode-verified against v2.4.10 (file:line references available on request).

### Why compiles were re-scanning the classpath every time

BTA 2.4.10 tears down the compiler **application environment after every compilation**: the `KotlinCoreEnvironment` ref-count drops to zero per exec, and even `kotlin.environment.keepalive` still clears the FastJarFileSystem handler caches via `idleCleanup()` (KotlinCoreEnvironment.kt:577-579). So every warm compile re-parsed the central directories of all ~1800 classpath jars and rebuilt their VFS trees — that was the 23% `jarfs.FastJar*` share in the JFR above, plus a good part of the FIR-resolution share.

### The fix: `CompilerEnvironmentPin`

Upstream Kotlin **master**'s BTA already fixes this with an `ApplicationEnvironmentPin` held per `BuildSession`. We front-run it for 2.4.10: one reflective call inside the isolated impl classloader — `KotlinCoreEnvironment.Companion.getOrCreateApplicationEnvironmentForProduction(disposable, CompilerConfiguration())` — holds the ref-count so per-exec disposal skips both `disposeApplicationEnvironment()` and `idleCleanup()`. `Disposer.dispose(disposable)` on session close releases everything. The class is documented to be **deleted** when `mcp.kotlinc.version` reaches the 2.5 line (the pin comes built in), and any reflective mismatch fails loudly at that exact moment.

Design points (from a 3× independent review round):
- **Lazy and IN_PROCESS-only**: daemon-only sessions never create a client-side environment (test-pinned).
- **Lease-scoped**: the pin shares the session-lease lifecycle — acquired before the first in-process compile, released after `BuildSession.close()` even if that throws, including the deferred-close-during-in-flight-compile path (test-pinned).
- **Transactional acquisition**: a half-acquired pin disposes its own disposable before rethrowing.
- Production (`CodeEvalManager`) switches **DAEMON → IN_PROCESS**: BTA 2.4.10's daemon policy leases + releases a daemon session around *every* operation and clears the daemon-side jar caches each time (CompileServiceImpl.kt:878-881) — the pin cannot help that path from the client. Gradle's KGP has the same limitation (with a TODO in their source wishing otherwise).

### Measured (probe on the 1789-jar IDEA classpath, control reproduced baseline within 3%)

| Configuration | warm steady-state | vs old kotlinc subprocess (3–6 s) |
|---|---|---|
| BTA as previously on this branch | 367 ms | 8–16× |
| **BTA + pin (this commit)** | **218 ms** | **14–28×** |

Pin acquisition costs ~100 ms once per session. Cold first compile (~2 s) is unchanged — that cost is environment bootstrap and happens once per project.

### Memory: verified, not guessed

A dedicated pass over the 2.4.10 sources + shipped bytecode (decompiled `ConcurrentFactoryMap`, `FastJarHandler`, `Disposer.ObjectTree`):

- **No premature GC**: the pinned cache chain is strong references end-to-end (`ConcurrentFactoryMap.createMap` → plain `ConcurrentHashMap`; zero Soft/WeakReferences in the FastJar path). The IDE's memory pressure cannot silently evict the caches — and there's no pressure valve either.
- **No growth per compilation**: nothing app-env-scoped is keyed by project or compilation; for a **constant classpath** the retained heap is flat after the first compile. Estimated **~275 MB (220–350 MB)** for a production-scale classpath (~1.1 M jar entries at ~250 B/entry). The one hazard is a *varying* jar classpath (each new jar path pins a tree until unpin) — our snippet classpath is the constant `ideClasspath()` and per-exec outputs never join it.
- **Lifecycle is test-pinned**: same environment instance across compiles (assertSame), environment **null after `close()`** — disposed, nothing retained (assertNull), fresh environment on reuse-after-close, deferred-close path included. `CodeEvalManager.dispose()` (project close) releases it all.
- Open-file handles are bounded (SLRU of 20+10 mmaps) regardless of jar count.

### `doSetup()` globals — audited

The relocated `IdeaStandaloneExecutionSetup.doSetup()` writes JVM-global system properties (property *names* are not relocated). Verified against the IDE sources: **BTA already ran it before every in-process compile** (3 call layers in the shipped jars) — the pin adds zero new exposure; every registry-backed value it writes equals the current IDE bundle default; `idea.plugins.compatible.build`/`idea.ignore.disabled.plugins` consumers cache at IDE startup; `java.awt.headless` is NOT touched (that's the standalone CLI main only). The one real wart — `idea.config.path=some/non/existent/path` written when unset — is neutralized by pre-seeding the property with the host's real `PathManager.getConfigPath()` before the session starts.

### Trade-off to be aware of

In-process compilation moves the compiler's working set (~0.3 GB retained + compile transients) into the IDE heap — that is the price of ~220 ms compiles in 2.4.10. Escape hatches: `CodeEvalManager.dispose()`/session close frees everything; a `LowMemoryWatcher → idleCleanup()` pressure valve is a possible follow-up if field reports want it; and the BTA 2.5-line upgrade should let us revisit the daemon policy with upstream's own pinning.

jonnyzzz:
Final numbers after the pin landed (commits `4ad7b1af` + `12fc13f5`): the KtBlock test matrix had been rebuilding the build session — and with it the pinned environment — once per generated test class (~98× per fork). One shared session per test JVM (matching production semantics) later, the identical clean-cache benchmark reads:

| 355 KtBlocks, IDEA stable, 8 forks, clean cache | wall | Σ test-seconds |
|---|---|---|
| main (kotlinc subprocess per block) | 401 s | 1186 s |
| BTA as originally submitted | 128 s | 523 s |
| **BTA + environment pin + one session per JVM** | **45 s** | **194 s** |

**~9× wall-clock on the matrix, 14–28× per production exec_code compile (218 ms warm).** 355/355 green. Repeat runs additionally short-circuit through the compile cache.

ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/execution/CodeEvalManager.kt
jonnyzzz: Resolved by going one step further: the working dir is gone entirely. It only ever held the BTA impl jars extracted from resources — after the rebase the jars ship as real files in the plugin distribution's `kotlinc/` folder (successor of the old kotlinc dist layout), and `KotlinBuildsSession` takes the jar paths explicitly. No extraction, no temp/cache dir, nothing to clean in `dispose()`. (By "StorageManager" I meant our own `StoragePaths`/`executionStorage`, but the question dissolved.) This also fixed two latent issues: `copyTo(overwrite=false)` bricked the service after an unclean IDE exit, and `defaultStdlibJar` re-ran the extraction.

ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/execution/CodeEvalManager.kt
jonnyzzz: Done on the branch: messages are recorded via `CompilerMessageRenderer`, persisted raw (non-remapped, incl. on timeout and zero-message failures) as `kotlin.txt` under the exec folder, and forwarded to idea.log through the `KotlinLogger` wrapper — with `error` deliberately demoted to `warn` (a failing agent-script compile is normal operation and must not file IDE fatal-error reports). Agent-visible locations are remapped to user-code coordinates via the new `LineMapping.remapLine` (BTA's structured locations never matched the old regex remap), and each error carries the offending source line — K2 messages like "Unresolved label." name no symbol, and our hint engine matches on the code text.

kotlin-cli/src/main/kotlin/com/jonnyzzz/mcpSteroid/koltinc/KotlinBuildsSession.kt
jonnyzzz: Great catch — and it went deeper. `executeOperation` is blocking and ran inline inside `withTimeout`, so `jvmOperation.cancel()` was dead code and the timeout never actually stopped the compiler. Restructured: the operation runs in a child `async(Dispatchers.IO)`; timeout/caller-cancel invoke the cooperative `cancel()`; and `OperationCancelledException` is translated to `CancellationException` *inside the child* — the placement matters: a non-CE child failure supersedes the caller's `TimeoutCancellationException` in `coroutineScope` and would break the timeout contract (three independent reviewers converged on exactly this). Timeout + close-during-compile regression tests added; session close is lease-counted so `dispose()` can't tear the session down under an in-flight compile.

kotlin-cli/src/main/kotlin/com/jonnyzzz/mcpSteroid/koltinc/KotlinBuildsSession.kt
jonnyzzz: Went one further to the repo idiom: `120.seconds`.

ij-plugin/src/test/kotlin/com/jonnyzzz/mcpSteroid/koltinc/KotlincCommandLineBuilderIntegrationTest.kt
jonnyzzz: Re-pinned against the real component in `KotlinBuildsSessionTest`: (a) spaces in every path — the old "JPA Model" case — as a two-stage compile where the spaced classpath entry is load-bearing, plus loading and invoking the produced jar (the old test's end-to-end verification); (b) a >40K-character, 600-entry classpath compile, per-OS on TC — the historically fragile layer. The argfile quoting tests died with the argfile itself: under BTA the arguments never touch an OS command line (in-process they stay an in-memory argument object into `K2JVMCompiler.exec`), so there is no quoting layer left to test. Bonus find from writing these: `List<Path> + somePath` in Kotlin resolves to `plus(Iterable)` because `Path : Iterable<Path>` — it appends the path's *segments* as classpath entries; documented in the test.

kotlin-cli/src/main/kotlin/com/jonnyzzz/mcpSteroid/koltinc/KotlinBuildsSession.kt
jonnyzzz: The question resolved itself out of existence: we removed the daemon flow entirely (single in-process path, no execution-policy toggle). Measurements drove it — BTA 2.4.10 leases + releases a daemon session (and clears the daemon-side jar caches) around every operation, so warm compiles couldn't get below ~370 ms on that path, while in-process with a session-pinned application environment (what upstream master's BTA now does itself) runs at ~218 ms. Since our client is the IDE — one long-lived session per project, compiles serialized — in-process is strictly simpler: no daemon lifetime tuning, no orphan JVMs, and two fewer jars in the distribution (`kotlin-compiler-runner`, `kotlin-daemon-client`; `kotlin-daemon-embeddable` must stay — `JvmCompilationOperationImpl`'s constructor links `CompileService.TargetPlatform` from daemon-common eagerly even in-process). We'd revisit the daemon when BTA can keep its caches across operations — the same classpath-reuse case you're collecting evidence for.

jonnyzzz:
## Hardening round: one flow, pruned bundle, stability suite — evidence and results

Commit `2172bd26`. Summary of what changed and the proof it's stable and safe.

### One implementation flow (no toggles)

The `CompilationExecutionPolicy` enum, the `executionPolicy` parameter and the daemon policy object are gone — in-process with the pinned environment is the only path. The daemon flow was strictly worse in 2.4.10 (its jar caches are cleared around every operation server-side) and existed only as an option nobody selected.

### The predicted resource change — it happened exactly as expected

Removing the daemon flow makes the two daemon-*client* jars dead weight. Verified three ways before touching Gradle: v2.4.10 source reading, bytecode disassembly, and a standalone classloader probe (jar-set A/B/C/D compile matrix). Key subtlety the probe caught: **`kotlin-daemon-embeddable` must stay** — `JvmCompilationOperationImpl`'s constructor eagerly initializes `targetPlatform = CompileService.TargetPlatform.JVM` from daemon-common even for pure in-process compiles (removing it fails with exactly that `ClassNotFoundException`). The build's own gate then fired as designed:

```
> Task :ij-plugin:verifyBundledLibraries FAILED
> Bundled kotlinc (BTA impl) jars mismatch!
  Missing:
    - kotlinc/kotlin-compiler-runner-2.4.10.jar
    - kotlinc/kotlin-daemon-client-2.4.10.jar
  Unexpected:
```

The exact-set pin now enumerates the 10-jar bundle (with the daemon-embeddable rationale in a comment for the next version bump).

### Stability suite (new `KotlinBuildsSessionStabilityTest`, ~13 s, CI-deterministic)

Structural leak probes over GC guesswork — the probes read the same 2.4.x internals the pin manages (documented for deletion together at the 2.5 line):

| Test | Proves | Result |
|---|---|---|
| 60 sequential compiles | one env instance for the whole session; env ref-count exactly 1 at rest; jar-handler cache flat; heap after GC | green — **heap growth 0 MB** (fail bound 200 MB) |
| 15 close()+recompile churn cycles | env disposed to null + ref-count 0 after *every* close; fresh env per cycle; instance stays usable | green |
| 10× 1 ms-timeout storm | cancelled operations leak no leases, don't tear down the pinned env, session compiles normally after | green |
| error storm (12 fail + 12 ok interleaved) | failures don't degrade the session; renderer isolation — no diagnostic bleed across compiles | green |
| close-vs-compile race fuzz (8 rounds, two interleavings) | the lease/pin machinery never leaks under racing close() | green |
| 2 000-function source | large-input compile inside the default timeout | green |

Deliberately NOT testing parallel `compileKotlin` calls: BTA doesn't document `BuildSession` thread-safety and production serializes compiles — asserting a property nothing relies on would only manufacture CI flakes.

### Validation battery

- `:kotlin-cli:test` — 64/64 green (the converted long-classpath test also got fast: no daemon JVM spawn).
- `:ij-plugin:test` — 268 green (1 pre-existing unrelated skip).
- `buildPlugin` + `verifyBundledLibraries` + `verifyClassFileVersions` — green (55 485 class files ≤ Java 21).
- Full IDEA-stable KtBlock matrix on the pruned bundle, clean compile cache — **355/355 green, 51 s wall** (vs 401 s on main).
- Windows (eugene-x220, real machine): **64/64 green** (`gradlew.bat :kotlin-cli:test`, JDK 25, BUILD SUCCESSFUL in 3m16s). The Windows-critical scenarios all ran natively: spaces-in-every-path (5.8 s), >40K-char classpath (5.7 s — in-process, no command-line limit exists to hit), pin lifecycle + close-race tests, and the full stability suite including the 60-compile flatness check (27.9 s). No file-lock, path-separator, or case-sensitivity issues observed.

Net code delta of the round: **−47 lines** of production/test code plus the two new test files.

ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/execution/CodeEvalManager.kt
```diff
@@ -34,11 +40,36 @@ inline val Project.codeEvalManager: CodeEvalManager get() = service()
 class CodeEvalManager(
     private val project: Project,
 ) : Disposable {
-    override fun dispose() = Unit
+    override fun dispose() {
+        kotlinBuildsSession.close()
```
jonnyzzz: it is much better to register the session object (each) via Disposer.registerDisposable(this) { ... }

ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/execution/CodeEvalManager.kt
```diff
@@ -34,11 +40,36 @@ inline val Project.codeEvalManager: CodeEvalManager get() = service()
...
+    // plugin classloader: KotlinBuildsSession loads these jars into its own
+    // isolated classloader.
+    private val kotlinBuildsSession = run {
+        // The relocated IdeaStandaloneExecutionSetup.doSetup() — which BTA runs before
+        // EVERY in-process compile — writes `idea.config.path=some/non/existent/path`
+        // when the property is unset (property names are JVM-global, not relocated;
+        // every other property it writes equals the IDE bundle default). Pre-seed it
+        // with the host IDE's real config path so third-party readers never observe
+        // the garbage value.
+        if (System.getProperty("idea.config.path") == null) {
```
jonnyzzz, line 66: route that one under some safe folder, never to IDE's own config folder (temp folder will work for now)

ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/execution/CodeEvalManager.kt
```diff
@@ -77,62 +108,92 @@ class CodeEvalManager(
...
-                    // compile, so when it IS present the agent must see it in the result.
-                    resultBuilder.logMessage("Compiler Output:\n$compilerOutput")
+            val compilerMessageRenderer = RecordingCompilerMessageRenderer()
+            val compileResult = try {
+                kotlinBuildsSession.compileKotlin(
+                    sources = listOf(inputKt),
+                    destinationDir = outputJar,
+                    compilerMessageRenderer = compilerMessageRenderer,
+                ) {
+                    if (extraParams.isNotEmpty()) {
```
jonnyzzz: always not empty, just simplify that

ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/execution/CodeEvalManager.kt
jonnyzzz: Done in 73ebfb40 — the session now closes via a child disposable registered on the service (`Disposer.register(this, Disposable { session.close() })`) right where the session is created; `dispose()` itself is empty.

ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/execution/CodeEvalManager.kt
jonnyzzz: Done in 73ebfb40 — the pre-seed now points at `$java.io.tmpdir/mcp-steroid-kotlin-config` (a dedicated path, nothing is created there; readers just see a sane absolute value instead of doSetup's relative garbage). The IDE's own config folder is never advertised.

ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/execution/CodeEvalManager.kt
jonnyzzz: Done in 73ebfb40 — guard dropped, `applyArgumentStrings(extraParams)` runs unconditionally (an empty token list on a fresh builder is a no-op anyway).

jonnyzzz:
## Review 2026-08-03

- **Conflict is trivial**: `mergeable=CONFLICTING` reduces to a single file — `prompts/src/main/prompts/lsp/completion.md`. Both sides independently replaced deprecated `CompletionContributor.forLanguage`; main's version (bb47b125 + b52a0134: baseLanguage-chain matching, null/empty-language handling, MetaLanguage note) is strictly more complete than this branch's hunk — resolve by taking main's side. No other overlap: main since merge-base 8f52118c touched none of CodeEvalManager.kt, kotlin-cli/ij-plugin/prompts build scripts, or KtBlockCompilationTestBase.
- **Dev-repo scoping** (kotlin-cli/build.gradle.kts:13, ij-plugin/build.gradle.kts:69, prompts/build.gradle.kts:13): `https://packages.jetbrains.team/maven/p/kt/dev` is added unfiltered and listed BEFORE mavenCentral, so every dependency in those three projects probes the dev repo first. Add a `content { includeGroup("org.jetbrains.kotlin") }` (or `includeVersionByRegex` for the RC builds) filter to scope it.
- **Release gate / owner decision**: the bundle pins `kotlin-build-tools-impl:2.4.20-RC-197` — a numbered pre-release from a dev repository. Shipping a Marketplace plugin embedding an RC compiler, with build reproducibility tied to dev-repo retention, needs an explicit call: merge on RC now vs. wait for 2.4.20 on Maven Central. This appears to be why the PR is still draft.
- **ALikhachev's CHANGES_REQUESTED is formally outstanding** but its substance is addressed: unhandled `OperationCancelledException` → now translated to CancellationException inside the child coroutine (KotlinBuildsSession.kt, preserving the TimeoutCancellationException contract); daemon-lifetime question → daemon flow deleted entirely (KT-88183). Needs a re-review/dismissal before merge.
- **No closing keyword wired**: `closingIssuesReferences` is empty. The PR resolves #207 in substance (7.8 s avg compile → ~0.2–0.4 s warm, measured + JFR-profiled) — add "Closes #207" to the body.
- Convention sweep clean: no `runCatching{}.onFailure`, no new `internal`, no empty catches (the `catch (_: TimeoutCancellationException)` persists messages + reports failure), no Java threading primitives in coroutine code. Lease-based deferred session close and fail-fast `DEFAULT_JVM_TARGET` derivation look correct; the stability suite asserts real invariants (single pinned env, ref-count 1 at rest, flat jar-handler map, heap-growth cap, lease-leak checks) — coverage is honest.

**Verdict: blocked** (draft by design). Content is merge-quality and tests are credible; the gates are (1) the RC-197-from-dev-repo vs. 2.4.20-GA decision, (2) formal re-review after the CHANGES_REQUESTED threads were addressed, (3) the one-file rebase taking main's completion.md.

jonnyzzz:
@ALikhachev could you please take another look?

All three requested changes are implemented:
- `OperationCancelledException` is translated inside the child coroutine, with timeout/storm regression coverage preserving `TimeoutCancellationException`.
- The default is `120.seconds`.
- The daemon path was removed; compilation is a single in-process flow, now using BTA `2.4.20-RC-197` and its upstream KT-87743 environment pin.

I also rebased onto current `main`, took the more complete `main` side of the completion-guide conflict, scoped the Kotlin dev repository to `org.jetbrains.kotlin`, and restored the compiler-diagnostic renderer that the history rebase exposed.

Fresh verification on head `97a423e0`: `:kotlin-cli:test`, bundled-library exact-set gate, class-file gate, and the generated IDEA completion KtBlock test all pass. The leak/stability suite remains green with 0 MB measured retained-heap growth.

ij-plugin/build.gradle.kts
```diff
@@ -66,6 +66,11 @@ val verifierIdeProduct: IdeProduct = when (targetIdeProduct) {
 }
 
 repositories {
+    maven("https://packages.jetbrains.team/maven/p/kt/dev") {
```
Tapchicoma, line 70: Generally, I would advise against using `dev` repo - artifacts there are retained only for some period and eventually removed. 

Please migrate to `2.4.20-RC` release once it is out (planned on 12.08) and stop using this repo.

kotlin-cli/src/main/kotlin/com/jonnyzzz/mcpSteroid/koltinc/KotlinBuildsSession.kt
```diff
@@ -0,0 +1,257 @@
...
+            }
+        } finally {
+            releaseSession(lease)
+        }
+    }
+
+    /**
+     * Closes the current build session managed by this instance.
+     *
+     * Thread-safe. In-flight compilations keep the session alive until they finish
```
Tapchicoma: I am curious what is the use-case for awaiting running compilations if `close()` was called? I expect when `close()` is called - caller is not interested in getting running compilation results.

ij-plugin/build.gradle.kts
jonnyzzz: Fixed in 4f092e7b798910d36078385017c17f2e0750c8d6 / #438 (stacked on this branch). Narrowed the kt/dev content filter to pre-release kotlin coordinates only — `includeVersionByRegex("org\\.jetbrains\\.kotlin", ".*", ".*-(dev|RC|Beta|M)[0-9]*(-[0-9]+)?")` instead of `includeGroup` — so kt/dev now serves only the seven 2.4.20-RC-197 modules; `kotlin-stdlib:2.3.20`, `kotlin-reflect:1.6.10` and every other release version resolve from Maven Central again. Traced with `--refresh-dependencies --info` across all three affected modules. Also moved `mavenCentral()` ahead of the kt/dev block in all three scripts, and consolidated the 13 hardcoded version literals into a single `mcp.kotlin.bta.version` gradle property.

Full removal of kt/dev plus the 2.4.20-RC bump is left as a TODO on each block for after ~12.08 once the RC is on Central — KT-87743 (ApplicationEnvironmentPin) is not in the Beta2 that is currently there, so bumping now would break the build. Thanks for the heads-up on retention: 2.4.20-RC-183 is already a 404.

kotlin-cli/src/main/kotlin/com/jonnyzzz/mcpSteroid/koltinc/KotlinBuildsSession.kt
jonnyzzz: `close()` never awaits a compilation — it is O(1) and non-blocking: it detaches the session under a lock and hands the underlying `BuildSession.close()` to whichever thread releases the last lease. What is deferred is teardown only: `clearJarCaches()` (the `ZipHandler` accessor cache and the jar-filesystem handlers), the `ExecutorService` that in-process operations are submitted to, and on 2.4.20 the KT-87743-pinned application environment. Firing that mid-compile would pull the compiler environment out from under a live resolve. Cancellation already happens one layer up (the `compileKotlin` timeout, or cancelling the calling coroutine), so the deferral only unwinds an operation that is already being cancelled. I reworded the KDoc to state exactly this — `close()` returns immediately and neither blocks nor cancels.

Your question did surface a real bug right next to it, though: there was no `closed` state, so a compile arriving *after* `close()` opened a brand-new `BuildSession` that nobody would ever close — production closes exactly once, from the `CodeEvalManager` Disposer child — leaking the KT-87743 environment pin, its jar caches and the isolated compiler classloader whenever `exec_code` raced project disposal. `close()` is now terminal and rejects further compiles under the same lock (the non-terminal "drop caches and keep compiling" behaviour moved to `dropCaches()`).

Fixed in #439 (`1f8e6cc`), stacked on this branch, with a race test that fails without the guard.

jonnyzzz: Follow-up in 88d5071a: after clarifying the actual lifecycle, I removed the lease/activeOperations/deferred-close machinery entirely.

KotlinBuildsSession now owns one BuildSession for its full lifetime; callers serialize compilation (CodeEvalManager already does this with compilationMutex), and close() is called only during disposal after the final compile. close() directly closes that BuildSession; close-vs-compile and reuse-after-close are no longer supported contracts.

I deliberately did not create/close a BuildSession per compilation: KT-87743 pins the compiler application environment and JarFS caches to the BuildSession lifetime. Per-compile sessions would throw away the cache responsible for the measured clean-matrix result (~54 s shared session versus ~128 s unpinned/per-session behavior).

Validation: :kotlin-cli:test 62/62, including the 60-compile cache/heap stability test and repeated fresh-session cleanup; a focused generated prompt compilation class also passed both variants; edited files are inspection-clean.

@Tapchicoma please take another look.

ij-plugin/build.gradle.kts
jonnyzzz: Re-checked the non-expiring Kotlin bootstrap repository on 2026-08-04 before changing this: 2.4.20-RC-197 is not published there (both BTA API and implementation POMs are 404), and every currently published bootstrap BTA implementation in the 2.4.20 and 2.5.0 lines lacks ApplicationEnvironmentPin/KT-87743.

So this branch still needs the scoped kt/dev repository for the fixed build. The BTA versions remain direct literals at each use site (no version catalog, property, or provider), and the project's Kotlin Gradle plugin remains 2.3.20. TODO-BTA-PERF.md now records the bootstrap audit and the required switch back to the first regular Maven Central Kotlin release containing KT-87743 (bootstrap is acceptable sooner if a verified fixed build appears).

@Tapchicoma this thread is intentionally waiting on publication; please let me know if you know a fixed bootstrap/Central coordinate that is already available.

jonnyzzz:
@Tapchicoma please take another look.

Latest cleanup after merging current `main`:

- Neither OCR nor kotlinc is part of the `npx-kt` fixture contract. The fixture now uses the neutral `bin/fixture-executable` path solely to verify that plugin deployment preserves Unix executable bits.
- `prompts` no longer declares the Kotlin dev repository or hardcodes `2.4.20-RC-197`. It consumes `:kotlin-cli` non-transitively plus that module's existing `kotlincDist` artifact; the cache key reads the compiler version from the loaded BTA.
- The project Kotlin Gradle plugin remains 2.3.20.
- `TODO-BTA-PERF.md` is now `docs/bta-snippet-compilation-performance.md`. Its only follow-ups are switching to the first regular Kotlin release containing KT-87743 and monitoring KT-88182, KT-88183, and KT-88278.
- The unrelated `de.undercouch.download` 5.7.0 bump is removed from this PR and isolated in #440.
- I rechecked the publication constraint: there is still no verified fixed Central/bootstrap BTA coordinate, so the scoped kt/dev repository remains only where `:kotlin-cli` / plugin packaging need it. That thread intentionally stays open until such a build exists.

Post-merge validation: Gradle sync; `:kotlin-cli:test` 62/62; focused generated KtBlock compilation 2/2; `PluginDeployTest` 2/2. The PR is mergeable again.

jonnyzzz:
@Tapchicoma @ALikhachev please take another look at the latest two commits.

The compiler-specific timeout/cancellation path has now been removed entirely: no `compilationTimeout`, `BuildOperation.cancel()`, `OperationCancelledException` translation, or timeout-only result handling/tests. BTA's blocking operation runs on `Dispatchers.IO` and completes normally.

I also tested the suggestion to create and close a BuildSession for every compilation. On the same 18 real IDEA/PyCharm/CLion classpath cases with a clean KtBlock cache:

- shared field-owned session: **15.023 s**
- fresh session per compilation: **24.587 s**

Fresh sessions are 1.64× slower because close releases KT-87743's application-environment/JarFS pin and the next session rebuilds the classpath caches, so the simple project-owned session remains and is disposed at project shutdown. The rationale and measurements are now in KDoc and `docs/bta-snippet-compilation-performance.md`.

Fresh verification: `:kotlin-cli:test` 60/60, prompt compilation 18/18, `:ij-plugin:compileKotlin` green, edited-file IntelliJ inspections clean. The upstream watch list now also includes KT-57097 alongside KT-88182, KT-88183, and KT-88278.

jonnyzzz:

Issue #362, opened 2026-07-28
JDK folder should not have devrig version


jonnyzzz:
We need to include JDK original verison, not Devrig version when install.ps1/install.sh are downloading the JDK.

<img width="349" height="110" alt="Image" src="https://github.com/user-attachments/assets/6c4e6e33-8691-4bc1-bc2d-7054d5f1c7a5" />

jonnyzzz:
Fixed in **v0.102**. Release notes: https://devrig.dev/releases/0.102/ · GitHub release: https://github.com/jonnyzzz/mcp-steroid/releases/tag/v0.102

Issue #228, opened 2026-07-03
Error resilience


BraindeadDogge:
- Disk-space precheck before a 500 MB download (fail early, clearly).
- Auto-retry with backoff on transient network failure insteaimmediately.

BraindeadDogge:
Implemented error resilience in the generated install.sh / install.ps1 (the ~611 MB devrig + JDK download path):

- Disk-space precheck before any download -- fails early with a clear need ~N MB, have ~M MB free message instead of dying mid-stream. Real per-platform sizes are now baked into the scripts by the generator
- Retry with backoff (3 attempts, growing delay) around fetch+verify, so a transient network failure or a corrupted (SHA-mismatch) transfer re-downloads; only the final attempt is fatal
- Network timeouts so retry can actually fire: curl --connect-timeout + stall detection (--speed-limit/--speed-time, which doesn't cap a slow-but-progressing download), wget --timeout; PowerShell uses a generous -TimeoutSec ceiling
- Staging cleanup on any exit, a broken download no longer leaks hundreds of MB

Covered by unit tests (baked sizes + resilience machinery + size validation) and two gated Docker scenarios (insufficient disk; 404 origin => retries then fails cleanly with staging removed).

BraindeadDogge:
Think about incremental download after fail

jonnyzzz:
Triage 2026-07-31: Correction to the 'implemented' comment above: the resilience work is in PR #363 (stack 1/6), which is still OPEN — main's install.sh.tmpl/install.ps1.tmpl have no disk precheck or retry today. Keep this open until #363 merges.

PR #363, opened 2026-07-28
Installers: report download progress and survive bad downloads (stack 1/6)

BraindeadDogge:
A ~611 MB install previously gave the user no feedback and no second chance: a single transient network blip or truncated transfer aborted the whole thing, and a full disk failed opaquely mid-stream.

- Bake each artifact's byte length into the scripts and announce it in a `downloading <kind> (~N MB) from <url>` line. **The wording is identical in `install.sh` and `install.ps1` on purpose** — the IDE plugin (PR 5) parses this line to render a real progress fraction.
- Retry download+verify as a unit with growing backoff, so a failed *or* sha-mismatched transfer re-downloads. Only the final attempt is fatal.
- Pre-check free disk space against `(devrig + jdk) × 3` and fail early with the numbers.
- Clean up this run's staging on failure, so a broken download cannot leak hundreds of megabytes.

### DIFF
--- a/installer-gen/src/main/resources/templates/install.ps1.tmpl
+++ b/installer-gen/src/main/resources/templates/install.ps1.tmpl
@@ -10,6 +10,10 @@
 
 Set-StrictMode -Version Latest
 $ErrorActionPreference = 'Stop'
+# Invoke-WebRequest renders a per-chunk progress bar by default, which is extremely slow on Windows
+# (repeated console redraws). The "downloading $kind (...)..." log line below is the textual progress
+# instead; disabling the built-in bar restores normal download throughput.
+$ProgressPreference = 'SilentlyContinue'
 [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
 # PowerShell's default progress bar redraws on every byte read (Invoke-WebRequest) and every entry
 # unpacked (Expand-Archive), turning a multi-hundred-MB download + JDK unpack into a UI-bound crawl
--- a/installer-gen/src/main/resources/templates/install.ps1.tmpl
+++ b/installer-gen/src/main/resources/templates/install.ps1.tmpl
@@ -28,6 +28,11 @@ $Version      = '@@VERSION@@'
 $DevrigUrl    = '@@DEVRIG_URL@@'
 $DevrigSha256 = '@@DEVRIG_SHA256@@'
 $DevrigFormat = '@@DEVRIG_FORMAT@@'
+# devrig-dist archive byte length - used (with the per-platform JDK size) for the pre-download
+# disk-space check, so a full disk fails early and clearly instead of mid-stream.
+$DevrigSize   = @@DEVRIG_SIZE@@
+# Download resilience (#228): re-download a failed OR corrupted (sha-mismatch) artifact this many times.
+$DlAttempts   = 3
 # The devrig launcher subpath INSIDE the unpacked dist (computed + asserted by the generator from the
 # real zip - `devrig-<version>-<hash>\bin\devrig.bat`). The script runs this and lets devrig register
 # itself (`devrig install devrig`), so the script never writes the launcher/PATH.
@@ -39,7 +44,17 @@ $Platforms = @{
 }
 
 function Write-Log([string]$m) { [Console]::Error.WriteLine("[mcp-steroid] $m") }
-function Fail([string]$m) { Write-Log "ERROR: $m"; exit 1 }
+# Remove THIS run's staging so a broken download never leaks hundreds of MB. Guard the $BinariesDir
+# lookup: Fail (which calls this) can run before $BinariesDir is defined, and StrictMode errors on an
+# undefined variable. PowerShell does not reliably run finally{} on `exit`, so cleanup lives here.
+function Remove-Staging {
+  $bd = Get-Variable -Name BinariesDir -ValueOnly -ErrorAction SilentlyContinue
+  if ($bd) {
+    Get-ChildItem -Path $bd -Filter ".tmp.$PID.*" -ErrorAction SilentlyContinue |
+      ForEach-Object { Remove-Item -Recurse -Force $_.FullName -ErrorAction SilentlyContinue }
+  }
+}
+function Fail([string]$m) { Remove-Staging; Write-Log "ERROR: $m"; exit 1 }
 
 # Home resolution: USERPROFILE -> HOME -> GetFolderPath (so pwsh-on-Linux tests can locate it too).
 $InstallRoot = $env:USERPROFILE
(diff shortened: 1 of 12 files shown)


installer-gen/src/main/resources/templates/install.ps1.tmpl
```diff
@@ -91,7 +106,19 @@ Get-ChildItem -Path $BinariesDir -Filter '.tmp.*' -ErrorAction SilentlyContinue
   Where-Object { $_.LastWriteTime -lt (Get-Date).AddDays(-1) } |
   ForEach-Object { Remove-Item -Recurse -Force $_.FullName -ErrorAction SilentlyContinue }
 
-function Install-Artifact([string]$kind, [string]$url, [string]$sha, [string]$fmt) {
+# -- preflight: disk space (#228) --
+# Fail early + clearly if the volume backing ~\.mcp-steroid cannot hold the download + unpack, rather
+# than dying mid-stream. Estimate the need as (devrig + jdk) archive bytes x3 (archive + unpacked tree
+# + slack).
+$needBytes = ([long]$DevrigSize + [long]$p.JdkSize) * 3
```
jonnyzzz: since we generate it, we know the real size of unpacked bianries, shall we use?

installer-gen/src/main/resources/templates/install.ps1.tmpl
```diff
@@ -103,11 +130,32 @@ function Install-Artifact([string]$kind, [string]$url, [string]$sha, [string]$fm
...
+  # Retry download+verify as a unit (transient network failure OR sha mismatch both re-download), with
+  # growing backoff; only the final failure is fatal. -TimeoutSec is a generous hard ceiling: PowerShell
+  # has no stall-only timeout (unlike curl's --speed-time), so a very slow link relies on this ceiling.
+  # install.sh is the primary macOS/Linux path.
+  $attempt = 1
+  $backoff = 2
+  while ($true) {
+    $err = $null
+    try {
+      Invoke-WebRequest -Uri $url -OutFile $tmpZip -UseBasicParsing -TimeoutSec 1800
```
jonnyzzz: why timeout here? Will it limit the download time?

installer-gen/src/main/resources/templates/install.sh.tmpl
```diff
@@ -143,20 +170,38 @@ main() {
   fi
 
   # Bind the concrete download + checksum tools now that the preflight proved one of each exists.
+  # Both carry timeouts so the retry loop above can actually fire: --connect-timeout caps a dead-host
+  # connect, and --speed-limit/--speed-time (curl) / --timeout (wget read timeout) abort a STALLED
+  # transfer without capping a legitimately slow -- but progressing -- ~600 MB download.
   if command -v curl >/dev/null 2>&1; then
-    fetch() { curl -fsSL "$1" -o "$2"; }
+    fetch() { curl -fL --progress-bar -S --connect-timeout 30 --speed-limit 1024 --speed-time 30 "$1" -o "$2"; }
```
jonnyzzz, line 182: what is --speed-limit?

installer-gen/src/main/resources/templates/install.sh.tmpl
```diff
@@ -43,9 +50,9 @@ promote_tree() {
   fi
 }
 
-# install_artifact <kind> <url> <sha256> <format>  -> prints the target dir on stdout
+# install_artifact <kind> <url> <sha256> <format> <size_bytes>  -> prints the target dir on stdout
 install_artifact() {
-  ia_kind="$1"; ia_url="$2"; ia_sha="$3"; ia_fmt="$4"
+  ia_kind="$1"; ia_url="$2"; ia_sha="$3"; ia_fmt="$4"; ia_size="${5:-0}"
```
jonnyzzz: why `:-0`, we control how each calls are done, no need to fallback.

jonnyzzz (changes requested):

installer-gen/src/main/resources/templates/install.sh.tmpl
BraindeadDogge: --speed-limit 1024 --speed-time 30 = "abort if the average transfer rate stays below 1024 bytes/s for 30 sec" These 2 flags work only in pairs, and it's very important in this PR, since otherwise the process would stay forever, never triggering a retry cycle

installer-gen/src/main/resources/templates/install.ps1.tmpl
BraindeadDogge: Agree, done in [446ccc5](https://github.com/jonnyzzz/mcp-steroid/pull/363/commits/446ccc56b6b40ba8c300620e299a9e4a79566a4e). 
As a matter of fact i also decided to compare previous and current result and here is what i got: 
LINUX-X64        archive=217 MB  unpacked=390 MB  ratio=1.79 | old(×3)=651 MB  new=667 MB  (+16)
LINUX-AARCH64    archive=214 MB  unpacked=389 MB  ratio=1.81 | old(×3)=644 MB  new=665 MB  (+21)
MACOS-AARCH64    archive=208 MB  unpacked=376 MB  ratio=1.80 | old(×3)=624 MB  new=643 MB  (+19)
WINDOWS-X64      archive=211 MB  unpacked=374 MB  ratio=1.76 | old(×3)=635 MB  new=644 MB  ( +9)
WINDOWS-AARCH64  archive=196 MB  unpacked=309 MB  ratio=1.57 | old(×3)=590 MB  new=556 MB  (−34)

So the exact numbers move the requirement by −30 to +21 MB, therefore x3 wasn't that bad guess, though now we don't depend on luck and future changes, thank you for noticing

installer-gen/src/main/resources/templates/install.ps1.tmpl
BraindeadDogge: Originally this solution was intended as a stall detection, though for big file as we have this timeout would require at least 30 Mbit/s of average speed, which might still be optimistical for some users, so here is what i've done in [4b96ec0](https://github.com/jonnyzzz/mcp-steroid/pull/363/commits/4b96ec0cc60a1c016b36d1ccabe6ace500e7da03):
- switched to curl and use the same stall detection i'm using for .sh (under the hood of Windows PowerShell 5.1 curl is the alias for Invoke-WebRequest, therefore i've also added a script to make sure we're using original curl)
- a fallback to Invoke-WebRequest though with updated time limits to 7200s(8 MBit/s on average)

jonnyzzz:
## Review 2026-08-03

- **Head does not compile (CI red).** Commit 446ccc5 made `unpackedSize` a required param on `JdkScriptEntry`/`DevrigEntry`, but `test-integration-agent-launch/.../InstallerPs1ExecutionTest.kt:106,107,115` still pass only `size` — `:test-integration-agent-launch:compileTestKotlin` fails on the windows-native job (run 30629533559: "No value passed for parameter 'unpackedSize'"). Same class of miss the a632817 reconcile commit fixed for the `size` field.
- **Second latent Windows failure behind the compile error.** `InstallerScriptTest.renderPs1()` never replaces `@@DEVRIG_UNPACKED_SIZE@@`; the template line `$DevrigUnpackedSize = @@DEVRIG_UNPACKED_SIZE@@` is unquoted, so the rendered script fails the `Parser::ParseFile` syntax test (splatting-token error). Its hand-built `$Platforms` table also lacks `JdkUnpackedSize`, which the disk precheck reads.
- **7-file merge conflict with main, and it's semantic, not textual.** `git merge-tree` conflicts on both templates, `InstallerGenerator.kt`, and all four test files. Root cause: main's 0c192251 (#362) also added a 5th positional arg to `install_artifact`/`Install-Artifact` (`ia_version="$5"`) plus a `JdkScriptEntry.version` field — this PR claims `$5` for `ia_size`. The rebase must renumber args across both templates, renderShCase/renderPsTable, validateScriptTable, and every fixture; #399/#398's rework of the same templates lands in the same hunks.
- Owner's requested changes are substantively addressed: exact `unpackedSize` replaces the ×3 guess (446ccc5), curl.exe with `--speed-limit/--speed-time` stall detection is the primary PS path with an Invoke-WebRequest fallback for pre-1803 (4b96ec0), and the `:-0` default is gone. The resilience design itself (retry+verify as a unit, disk precheck, per-run staging cleanup trap) fits the #399 contract (install+PATH only) and resolves both halves of #228.
- Note for the stack: `downloading <kind> (~N MB) from <url>` is the cross-PR contract line the IDE plugin (stack 5/6) parses; the generator test pins both script shapes — keep that test in sync during the rebase.

**Verdict: needs-changes** — the content is on the right track and the review feedback was addressed, but the branch head is red on its own (compile break + unreplaced ps1 placeholder in the parse test), and the stale-base conflict with #362's identical-arity signature change requires a real reconcile, which then ripples through stack PRs 2–6.

Issue #284, opened 2026-07-16
devrig CLI: derive command registration, flags, and help from ToolSpec


matveidetonator:
Follow-up to PR #272 review comment C8.

Goal: make a newly added MCP tool automatically, or with minimal explicit glue, available as a `devrig` CLI command.

Explore deriving from each tool's `ToolSpec` / `InputSchemaElement`:

- CLI command registration
- option names, required/optional status, and value types
- concise per-command help
- argument mapping into `ToolCallContext`

Constraints:

- keep the MCP tool surface unchanged
- preserve the existing `--json` envelope and exit-code contract
- allow explicit CLI-only affordances where needed (`--out`, `--wait`, stdin/file input)
- avoid copying validation and parameter-building logic already owned by `ToolSpec.call()`

PR #364, opened 2026-07-28
Auto-detect the target project from the agent's working directory (stack 2/6)

BraindeadDogge:
Every devrig tool call required an explicit `project_name`, which the agent had to fetch from `steroid_list_projects` first — an extra round trip on every call, and a routing key it frequently guessed wrong.

`project_name` becomes **optional on the devrig surface**: when omitted, the project owning the agent's working directory is used. The in-IDE surface is unchanged.

- **`DevrigWorkspaceRoots`** — candidate directories from the MCP client's advertised roots, queried with a short timeout so a slow or absent reply degrades to the process cwd rather than hanging.
- **`DevrigProjectDetection`** — matches a candidate directory against open projects, including when the cwd is a *subdirectory* of the project root.
- **`DevrigProjectResolver`** — joins the two and reports actionable errors: nothing discovered, no match, ambiguous match.
- **`markRequired()`** — lets one schema element be emitted as required in-IDE and optional on the devrig surface, so relaxing it doesn't fork the schema.

### DIFF
--- a/mcp-steroid-server/src/main/kotlin/com/jonnyzzz/mcpSteroid/server/ExecuteCodeTool.kt
+++ b/mcp-steroid-server/src/main/kotlin/com/jonnyzzz/mcpSteroid/server/ExecuteCodeTool.kt
@@ -75,11 +75,14 @@ data class ExecCodeParams(
 /**
  * Handler for the steroid_execute_code MCP tool.
  */
-class ExecuteCodeToolSpec(val handler: () -> ExecuteCodeToolHandler) : McpToolBase() {
+class ExecuteCodeToolSpec(
+    val projectNameRequired: Boolean = true,
+    val handler: () -> ExecuteCodeToolHandler,
+) : McpToolBase() {
     override val name = "steroid_execute_code"
     override val description get() = ExecuteCodeToolDescriptionPromptArticle().readPayload(PromptsContext.Generic)
 
-    val projectName = CommonToolParams.projectName().registerToSchema()
+    val projectName = CommonToolParams.projectName(projectNameRequired).registerToSchema()
 
     val code = InputSchemaElement.param("code")
         .description("Kotlin suspend method body")
@@ -123,7 +126,7 @@ class ExecuteCodeToolSpec(val handler: () -> ExecuteCodeToolHandler) : McpToolBa
         .registerToSchema()
 
     override suspend fun call(context: ToolCallContext): ToolCallResult {
-        val projectName = context[projectName]
+        val projectName = context[projectName].orEmpty()
         val code = context[code]
         val taskId = context[taskId]
         val reason = context[reason]
--- a/mcp-steroid-server/src/main/kotlin/com/jonnyzzz/mcpSteroid/server/VisionInputTool.kt
+++ b/mcp-steroid-server/src/main/kotlin/com/jonnyzzz/mcpSteroid/server/VisionInputTool.kt
@@ -17,7 +17,10 @@ import kotlinx.serialization.Serializable
 /**
  * Handler for the steroid_input MCP tool.
  */
-class VisionInputToolSpec(val handler: () -> VisionInputToolHandler) : McpToolBase() {
+class VisionInputToolSpec(
+    val projectNameRequired: Boolean = true,
+    val handler: () -> VisionInputToolHandler,
+) : McpToolBase() {
     override val name = "steroid_input"
 
     override val description = """
@@ -43,7 +46,7 @@ class VisionInputToolSpec(val handler: () -> VisionInputToolHandler) : McpToolBa
         Click coordinates with the screenshot target (e.g. @120,200) are interpreted relative to the window as reported by steroid_list_windows / steroid_take_screenshot.
     """.trimIndent()
 
-    val projectName = CommonToolParams.projectName().registerToSchema()
+    val projectName = CommonToolParams.projectName(projectNameRequired).registerToSchema()
 
     val taskId = CommonToolParams.taskId().registerToSchema()
 
@@ -60,7 +63,7 @@ class VisionInputToolSpec(val handler: () -> VisionInputToolHandler) : McpToolBa
         .registerToSchema()
 
     override suspend fun call(context: ToolCallContext): ToolCallResult {
-        val projectName = context[projectName]
+        val projectName = context[projectName].orEmpty()
         val taskId = context[taskId]
         val reason = context[reason]
         val windowId = context[windowId]
--- a/npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/DevrigServices.kt
+++ b/npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/DevrigServices.kt
@@ -4,7 +4,11 @@ import com.jonnyzzz.mcpSteroid.devrig.monitor.IdePidDiscoveryService
 import com.jonnyzzz.mcpSteroid.devrig.monitor.IdeProjectMonitorService
 import com.jonnyzzz.mcpSteroid.devrig.monitor.IntelliJPortDiscovery
 import com.jonnyzzz.mcpSteroid.devrig.server.DevrigBackendService
+import com.jonnyzzz.mcpSteroid.devrig.server.DevrigProjectResolver
 import com.jonnyzzz.mcpSteroid.devrig.server.DevrigProjectRoutingService
+import com.jonnyzzz.mcpSteroid.devrig.server.DevrigWorkspaceRoots
+import com.jonnyzzz.mcpSteroid.mcp.McpRootsService
+import com.jonnyzzz.mcpSteroid.mcp.McpSession
 import com.jonnyzzz.mcpSteroid.server.NPX_STREAM_IDLE_TIMEOUT_MILLIS
 import com.jonnyzzz.mcpSteroid.testHelper.CloseableStack
 import io.ktor.client.HttpClient
@@ -17,6 +21,7 @@ import kotlinx.coroutines.CoroutineName
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.withTimeoutOrNull
 
 class DevrigServices(
     private val lifetime: CloseableStack,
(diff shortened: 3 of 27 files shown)


mcp-steroid-server/src/main/kotlin/com/jonnyzzz/mcpSteroid/server/ExecuteCodeTool.kt
```diff
@@ -75,11 +75,14 @@ data class ExecCodeParams(
...
-class ExecuteCodeToolSpec(val handler: () -> ExecuteCodeToolHandler) : McpToolBase() {
+class ExecuteCodeToolSpec(
+    val projectNameRequired: Boolean = true,
+    val handler: () -> ExecuteCodeToolHandler,
+) : McpToolBase() {
     override val name = "steroid_execute_code"
     override val description get() = ExecuteCodeToolDescriptionPromptArticle().readPayload(PromptsContext.Generic)
 
-    val projectName = CommonToolParams.projectName().registerToSchema()
+    val projectName = CommonToolParams.projectName(projectNameRequired).registerToSchema()
```
jonnyzzz, line 85: Form the standpoint of classical OOP -- you would require the ProjectName factory in the contstructor and the factory itself encapsulate the secret on if project_name is required or not.

jonnyzzz:
What is the commandline at which agent starts the MCP process? How is that documented for each of the agents?

jonnyzzz:
This PR does not depend from the step1, so you can rebase it to `main`

jonnyzzz:
A potential good pre-set PR -- introduce project_name factory, pass it to all the constructors. Step 2 -- make the project_name factory for devrig side that creates non-required project_name as you do.

jonnyzzz:
This change will require to update Promts to explain an agent that project_name is now getting optional for devrig mode, that will require you to create the macro for promts to generate different content depending from the scenario.

jonnyzzz:
It feels like an overlap with @matveidetonator who was going to implement the same hehaviour for the CLI commands.

mcp-steroid-server/src/main/kotlin/com/jonnyzzz/mcpSteroid/server/VisionInputTool.kt
```diff
@@ -60,7 +63,7 @@ class VisionInputToolSpec(val handler: () -> VisionInputToolHandler) : McpToolBa
         .registerToSchema()
 
     override suspend fun call(context: ToolCallContext): ToolCallResult {
-        val projectName = context[projectName]
+        val projectName = context[projectName].orEmpty()
```
jonnyzzz, line 66: what does mean orEmpty() and why we need that at most of the places? 

Form my OOP standpoint, since we have the CommonToolParams factory, we may inject the logic into the factory itself. 

You either introduce a nullable project_name (which is not nice in the API + a breaking change), OR use another approach. Using "" there is not correct. 

I believe that we should better inject that resolution logic into the factory (that is also not ideal, because now spec introduce a non-trivial behavoir). 

From the API design standpoint, it's a good excersice to find a better approach to satisfy all requirements. We have the resolver already, which may turn the project_name (now nullable) to a project_name parameter. That may work.

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/server/DevrigExecuteCodeToolHandler.kt
```diff
@@ -10,15 +10,15 @@ import kotlinx.serialization.json.put
...
+    private val resolver: DevrigProjectResolver,
     private val beacon: DevrigBeacon,
 ) : ExecuteCodeToolHandler {
     override suspend fun executeCode(
         projectName: String,
         execCodeParams: ExecCodeParams,
         callProgress: McpProgressReporter,
     ): ToolCallResult {
-        val route = routing.requireProject(projectName)
+        val route = resolver.resolve(projectName)
```
jonnyzzz, line 21: To highlight the idea from the previous comment -- here you may have project_name inside the route, which is the non-null resolved project name. Thus a small change (and dedicated PR) to start usnig it at all places.

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/server/DevrigExecuteFeedbackToolHandler.kt
```diff
@@ -8,11 +8,11 @@ import kotlinx.serialization.json.put
 
 class DevrigExecuteFeedbackToolHandler(
     private val bridge: DevrigToolBridgeClient,
-    private val routing: DevrigProjectRoutingService,
+    private val resolver: DevrigProjectResolver,
     private val beacon: DevrigBeacon,
 ) : ExecuteFeedbackToolHandler {
     override suspend fun handleFeedback(projectName: String, params: FeedbackParams): ToolCallResult {
-        val route = routing.requireProject(projectName)
+        val route = resolver.resolve(projectName)
```
jonnyzzz, line 15: It is not clear why we need new service here, we can keep the existing one and chage it's internal logic

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/server/DevrigExecuteCodeToolHandler.kt
jonnyzzz: and if you follow that way, we can also introduce ProjectName data object, and make sure it's only typed object everywhere. 

ProjectNameRequest -> ProjectName data classes (carefull to keep the devrig -> mcp steroid interop

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/server/DevrigProjectDetection.kt
```diff
@@ -0,0 +1,46 @@
...
+    data object NoBackends : ProjectDetection
+}
+
+/**
+ * Selects the project a working directory belongs to. A route matches when its (canonical) project path
+ * equals or is an ancestor of a candidate dir (the agent may run from a subdirectory). Among matches,
+ * the deepest project in a single ancestor chain wins; two matches that are not in an ancestor relation
+ * are [ProjectDetection.Ambiguous].
+ */
+fun selectProjectByCwd(routes: List<ProjectRoute>, dirs: List<Path>): ProjectDetection {
```
jonnyzzz, line 20: There is only one CWD. Do we need many-to-many?

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/server/DevrigProjectResolver.kt
```diff
@@ -0,0 +1,39 @@
...
+import com.jonnyzzz.mcpSteroid.mcp.ToolCallErrorException
+
+/**
+ * Resolves the [ProjectRoute] for a project-scoped devrig tool call. An explicit, non-blank
+ * `project_name` always wins (delegates to [DevrigProjectRoutingService.requireProject]); otherwise the
+ * project is auto-detected from the agent's working directory. Detection failures surface as
+ * [ToolCallErrorException] with an actionable, case-specific message (task #226).
+ */
+class DevrigProjectResolver(
+    private val routing: DevrigProjectRoutingService,
```
jonnyzzz, line 13: it feels like a part of DevrigProjectRoutingService

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/server/DevrigProjectRoutingService.kt
```diff
@@ -43,6 +43,12 @@ class DevrigProjectRoutingService(
     fun requireProject(exposedProjectName: String): ProjectRoute =
         routeProject(exposedProjectName) ?: throw ProjectRouteNotFoundException(exposedProjectName)
 
+    /**
+     * Detects which open project the agent's working directory belongs to. See [selectProjectByCwd].
+     * Used by devrig to resolve a tool call that omits `project_name` (task #226).
+     */
+    fun detectProject(dirs: List<Path>): ProjectDetection = selectProjectByCwd(routes(), dirs)
```
jonnyzzz, line 50: every call to `routes()` is taking time, ideal approach is to have only 1 call per request

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/server/DevrigWorkspaceRoots.kt
```diff
@@ -0,0 +1,67 @@
...
+ */
+class DevrigWorkspaceRoots(
+    private val rootsProvider: suspend () -> List<Root>?,
+    private val processCwd: () -> String = { System.getProperty("user.dir") },
+) {
+    private val log = thisLogger()
+    private val mutex = Mutex()
+
+    @Volatile
+    private var cached: List<Path>? = null
```
jonnyzzz, line 27: why do you need mutex and caches? We may want coroutine-natural mutex. 

Projects roots can change over time, this invalidation of routes can be a problem. 

What performance problem are you solving here?

npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/DevrigServices.kt
```diff
@@ -90,6 +95,30 @@ class DevrigServices(
         )
     }
 
+    /** The single MCP client session for this devrig stdio process; set at initialize. */
+    @Volatile
+    var mcpClientSession: McpSession? = null
```
jonnyzzz, line 100: DevrigServices is the container of IoC components (we are not using IoC library, just declare all components here), having mutability here (beyond lazy initialization of services), does not feel correct.

jonnyzzz (changes requested):

jonnyzzz:
The main questions:
- does it make sense only for the CLI mode, not for the devrig MCP
- it depends how agent specify the working directory of MCP processes it starts
- it can be independent PR based/targeted to main

jonnyzzz:
**Review 2026-08-03**

- **Stale vs main — rework, not a mechanical rebase.** The 2026-07-28 reconcile predates f33da12b (devrig now registers via `McpSteroidTools.commonToolSpecs()`/`devrigToolSpecs()`, shared with the schema-driven CLI from the #284/#356 track) and 699ee376 (rewrote the `initialize`/`onSessionInitialized` block this PR edits in `StubStdioMcpServer.kt`). `registerAll(server, projectNameRequired = false)` has no equivalent on today's main, and threading the boolean through the five spec constructors would now also change the generated CLI surface — the exact overlap flagged in the 2026-07-30 comments. Note `mergeStateStatus: CLEAN` is vs `stack/1-installer-progress`, which is itself DIRTY vs main.
- **In-IDE surface regression, contradicting the PR body.** `markRequired()` keeps the parsed type nullable and handlers switch to `context[projectName].orEmpty()` (mcp-steroid-server/.../ExecuteCodeTool.kt:129, VisionInputTool.kt, VisionScreenshotTool.kt, FetchResourceToolHandler.kt). Previously `.required()` enforced at parse time (`ToolCallErrorException("Parameter project_name ... is required")`, mcp-core/.../McpSchema.kt:339–350). Now an in-IDE call omitting `project_name` silently routes `""` into the handler and fails with a confusing downstream error; only `ExecuteFeedbackTool` kept a blank check. Keep parse-time enforcement when `requiredInSchema = true`.
- **Prompt corpus untouched.** The exec_code tool description comes from `ExecuteCodeToolDescriptionPromptArticle` and the corpus still teaches fetching `project_name` from `steroid_list_projects` first — agents keep paying the round trip this PR exists to remove. Matches the earlier ask for a devrig/in-IDE scenario macro.
- Minor: `DevrigWorkspaceRoots.invalidate()` is never called (no `roots/list_changed` wiring), so roots changes are ignored for the process lifetime; and `selectProjectByCwd`'s `distinctBy { it.projectPath }` silently picks an arbitrary backend when two IDEs have the same project path open — arguably an Ambiguous case.
- Worth salvaging: the detection core is sound and well unit-tested (subdirectory match, deepest-leaf-wins, ambiguity), and the roots-with-timeout + cwd-fallback shape in `DevrigWorkspaceRoots` is right.

**Verdict: needs-changes** — the 2026-07-30 CHANGES_REQUESTED is unaddressed (no commits since 2026-07-28), and main's devrigToolSpecs()/CLI-metadata refactor invalidates the registration mechanism. Recommend: retarget to main as an independent PR, plug detection into `commonToolSpecs()` via a project_name factory, restore in-IDE required enforcement, and pair with the prompt-macro update.

Issue #253, opened 2026-07-08
[claude-plugin] Tested cross-platform script-only agent integration (MCP + hooks call installed devrig)


jonnyzzz:
## Goal

Find a **tested, cross-platform, script-only** way for a Claude Code plugin to integrate with the agent —
its **stdio MCP server** and its **hooks** — so they **call the already-installed `devrig`** on
macOS/Windows/Linux (and promote installing it), with **minimal dependencies** (no bundled native binary,
no assumed language runtime) so the solution is as robust as possible. Everything below is validated on
real Windows **and** Linux, on **both** CI systems (GitHub Actions + TeamCity), via the
`:test-integration-agent-launch` probe suite (mock scripts, no real devrig).

## Confirmed cross-OS matrix

Probe log line per launch: `LAUNCHED tag=<tag> via=<cmd|sh|shell|script>`.

| Launch form | Windows | Linux | Verdict |
|---|:-:|:-:|---|
| **MCP** — extensionless `command` | ✅ resolves to `.cmd` (cross-spawn/PATHEXT) | ✅ runs the `#!/bin/sh` file (execve) | **use this** |
| **MCP** — bare command name | ❌ | ❌ | plugin `bin/` not on subprocess PATH; also hijack risk |
| **Hook** — exec-form, extensionless path | ❌ | ✅ (`+x` shebang script) | not portable |
| **Hook** — exec-form, `.cmd` path | ❌ | ❌ | never |
| **Hook** — exec-form, `cmd /c <.cmd>` | ✅ | ❌ (no `cmd`) | Windows-only |
| **Hook** — shell-form, default shell | ✅ (Git Bash) | ✅ (`sh`) | portable |
| **Hook** — shell-form, `"shell":"powershell"` | ✅ | pwsh-dependent | Windows-native |
| **Hook** — shell-form → `#!/bin/sh` script (**devrig pattern**) | ✅ (Git Bash) | ✅ | **use this** |

Key: **hooks DO run in `-p` on Windows** (an earlier claim of mine to the contrary was a probe-form bug,
now corrected). devrig's shipped hooks are already in the working shell-form → they work on Windows **via
Git Bash**.

### MCP server — one command, two sibling scripts, no native binary
`.mcp.json`:
```jsonc
{ "mcpServers": { "devrig": { "type": "stdio",
    "command": "${CLAUDE_PLUGIN_ROOT}/bin/devrig-mcp" } } }   // NOTE: no extension
```
[...]

PR #365, opened 2026-07-28
Add `devrig connect` to wire an agent and an IDE together (stack 3/6)

BraindeadDogge:
Connecting an agent to MCP Steroid was a manual, error-prone sequence: register an MCP server by hand, find the plugin, install it into the right IDE. Two subcommands automate the two directions.

- **`devrig connect claude`** — enables the marketplace plugin through an *idempotent* `~/.claude/settings.json` merge, so re-running is a no-op.
- **`devrig connect ide`** — discovers running IDEs over their live ports and asks one to install the plugin, cascading **install dialog → browser → manual instructions** so the user is never left at a dead end. Skips already-connected IDEs, and returns a distinct `NO_IDE` exit code so callers can tell *failed* from *nothing to do*.
- The canonical MCP server name becomes **`devrig`**, with `mcp-steroid` retained as the legacy name so existing registrations are recognized and consolidated rather than duplicated.
- MCP command comparison is normalized before the canonical check, so Windows quoting differences no longer look like drift and trigger a redundant re-registration.

### DIFF
--- a/ai-agents/src/main/kotlin/com/jonnyzzz/mcpSteroid/aiAgents/McpServerConfig.kt
+++ b/ai-agents/src/main/kotlin/com/jonnyzzz/mcpSteroid/aiAgents/McpServerConfig.kt
@@ -3,7 +3,7 @@ package com.jonnyzzz.mcpSteroid.aiAgents
 
 import kotlin.collections.plus
 
-const val DEFAULT_SERVER_NAME = "mcp-steroid"
+const val DEFAULT_SERVER_NAME = "devrig"
 
 /**
  * Structured model of the MCP server connection info.
--- a/npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/ClaudePluginConnect.kt
+++ b/npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/ClaudePluginConnect.kt
@@ -1,0 +1,73 @@
+/* Copyright 2025-2026 Eugene Petrenko (mcp@jonnyzzz.com); Copyright 2025-2026 JetBrains. Use of this source code is governed by the Apache 2.0 license. */
+package com.jonnyzzz.mcpSteroid.devrig
+
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.JsonPrimitive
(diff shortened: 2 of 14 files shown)


jonnyzzz:
## Review 2026-08-03

- **`connect ide` is superseded by `devrig install plugin` (#391, merged to main 2026-07-30, after this stack's reconcile).** Main's `InstallPluginCommand.kt` already drives `/api/installPlugin` with a localhost Origin, and is more complete than `ConnectIde.kt`: per-IDE outcome taxonomy, `--check` mode, real compat gating (this PR's ConnectIde.kt:66-72 ignores the checkCompatibility body and "offers install anyway"), widened port scan, marker cross-referencing. Drop this half or rebuild it as a thin alias over the shipped orchestrator.
- **Hidden compile conflict on rebase**: `ConnectIde.kt:8` declares top-level `const val MCP_STEROID_PLUGIN_ID` — the identical declaration already exists in main's `InstallPluginCommand.kt:28`, same package. Both are new-vs-modified files, so git merges cleanly and the build then fails with conflicting top-level declarations.
- **Still novel and worth landing** (none of it is on main): `McpServerListReview.kt` `isPluginProvided()` + the install/check plugin-provided handling (#253, still open), `mcpCommandLinesEquivalent` Windows quote normalization (#160 — main's `InstallCommand.kt:382` still exact-compares `commandLine`, so the false-drift bug is live), and the `mcp-steroid`→`devrig` canonical-name swap (owner naming decision; also intersects the #399 CLI-surface contract — consider whether `connect claude` belongs under the `install` family instead of a new verb).
- **Sequencing hazard for `connect claude`**: it enables `devrig@jonnyzzz` sourced from `jonnyzzz/mcp-steroid`, but `.claude-plugin/marketplace.json` exists only on stack/4 (#366). Claude resolves the marketplace from the repo's live default branch, so running `devrig connect claude` after stack-3 merges but before #366 lands writes an entry pointing at a marketplace that does not exist. This half is effectively coupled to #366.
- **Stack staleness**: this PR reads CLEAN/MERGEABLE only against `stack/2-cwd-autodetect`; the stack's merge-base with main is 244b90c1 (2026-07-28) and main has moved 118 commits since — stack root #363 is already CONFLICTING. A full stack rebase is needed regardless.

**Verdict: needs-changes** — the `connect ide` half duplicates shipped #391 (and won't compile after rebase); the `connect claude` / #253 / #160 / rename parts are good and should be re-cut on top of current main, with `connect claude` sequenced with #366.

Issue #227, opened 2026-07-03
First-run discoverability


BraindeadDogge:
/devrig:help or a one-time welcome with copy-paste example prompts ("run the tests in the open IDE", "find duplicates in this file") so the user knows what the whole-IDE bridge can do.

BraindeadDogge:
Please note hook at the start of the session, this message appears only once after starting a new session with freshly installed devrig. 

<img width="1177" height="1120" alt="Image" src="https://github.com/user-attachments/assets/f31d8ddb-c03f-4d7a-ad69-3a94d7b1a5e3" />

jonnyzzz:
Reopened during the 2026-07-16 main-state audit. The described implementation exists only in open PR #140 (claude-plugin), which is not merged and currently conflicts with main. Keep this issue open under the Claude-plugin UX epic #222 until the implementation lands on main or the scope is explicitly rejected.

jonnyzzz:
Triage 2026-07-31: moved to 30-day growth: 50k vDAI breakout. Child of claude-plugin epic #222 (growth); ships with PR #140's help.md + SessionStart welcome.

PR #366, opened 2026-07-28
Add the devrig Claude Code plugin (stack 4/6)

BraindeadDogge:
A Claude Code marketplace plugin, so a user can reach the IDE without hand-editing any MCP configuration.

- Ships `.mcp.json` plus a polyglot launcher that `exec`s the installed devrig. The launcher is **script-only** and writes nothing to stdout before `exec` — stdout is the JSON-RPC channel, and one stray byte corrupts the protocol.
- `/devrig:setup`, `:status`, `:help`, `:uninstall` commands, plus hooks that report install state, nudge setup when devrig is missing, and surface recoverable errors where the agent can act on them.
- The build **stages, packages and validates**: it asserts the launcher set, the install path, marketplace name consistency and the command manifest, so a packaging mistake fails the build instead of reaching a user.


jonnyzzz:
## Review 2026-08-03

- **`:claude-plugin:check` fails on this branch as committed.** `claude-plugin/.claude-plugin/plugin.json` pins `"version": "0.100.0"` but the branch's own root `VERSION` is `0.101`; `validatePluginJson` (claude-plugin/build.gradle.kts) requires `0.101.0` and throws. Run `./gradlew :claude-plugin:syncClaudePluginVersion` and commit. CI did not catch this because only "Installer PS1 Verify" ran — the new `ciDistributionChecks` aggregator (root build.gradle.kts:356) is not wired into any GH workflow or TC step, so the validation gate this PR builds never executes on CI. Wire it (TC DSL repo or a workflow) as part of this PR series.
- **Windows never gets the IDE offer, and the marker is burned anyway.** claude-plugin/bin/offer-ide:18–23: when only `devrig.cmd` exists (every Windows install), the script skips `connect ide` entirely but takes `code=0` → burns the once-per-machine `ide-offered` marker and emits the "offer was made" nudge. That contradicts the script's own "burn ONLY for outcomes devrig actually reported" rule, and `validateOfferIdeRuns` has no cmd-only-launcher case to catch it.
- **Semantic staleness vs main.** `bin/offer-ide` hardcodes `devrig connect ide` (stack-3 `ConnectIde.kt`), but main has since shipped `devrig install plugin` (#391) for the same capability — the head branch has no `InstallPluginCommand`, so the rebase must reconcile the two. Bigger: #399 locked the contract "never auto-install the MCP Steroid plugin into IDEs — only via the explicit `devrig install plugin`, run by the user"; a SessionStart hook that auto-invokes an IDE install offer (may pop an IDE dialog / open a browser) needs an explicit owner decision against that contract and the #413 auto-registration policy before landing.
- Minor: claude-plugin/bin/install-devrig:28–33 captures the pipeline status of `curl … | sh`, i.e. sh's exit. A dead curl gives sh EOF → exit 0; with a pre-existing install the wrapper then reports success and clears the failure marker without having installed anything (the post-check only saves the fresh-install case). Fetch to a temp file or check curl's status separately.
- Housekeeping: no `Closes #…` keywords although this delivers #227/#242/#245/#246/#247/#253 — add them so the merge closes the wave. #140 (claude-plugin) is fully superseded by this PR (same scope, older bootstrap-based approach) and should be closed pointing here.

**Verdict: needs-changes** — the stale plugin.json version breaks the PR's own check gate, the Windows marker-burn is a concrete defect, and the `connect ide` vs `install plugin`/#399-contract reconcile is unresolved; also stack-blocked (#363 → #364 → #365 must merge first; delta reviewed vs stack/3 is otherwise solid, with genuinely strong behavioral validators).

Issue #360, opened 2026-07-27
Release and CI version strings are not comparable, so an IDE can treat either one as an update


CrazyCoder:
`build.gradle.kts` emits two differently shaped version strings:

```kotlin
// build.gradle.kts:95-99
version = when {
    isReleaseBuild -> "$baseVersion-$gitHash"
    providedBuildVersion != null -> providedBuildVersion   // "<base>.<counter>-(gh|jb)-<hash>"
    else -> "$baseVersion.$localBuildCounter-$gitHash"
}
```

A release is `0.101-40690055`. A CI build is `0.101.566-jb-90ac87b`. IntelliJ compares plugin versions with `VersionComparatorUtil`, through `PluginDownloader.compareVersionsSkipBrokenAndIncompatible`. That splits on `.` and `-`, splits each token again at digit and non-digit boundaries, and compares position by position. Numeric tokens outrank word tokens, and numbers compare by digit count first.

The third position therefore pits the leading digits of the release git hash against the CI counter. Running the real comparator out of `util_rt.jar`:

```
0.101-40690055   >  0.101.566-jb-90ac87b
0.101-4069005    >  0.101.566-jb-90ac87b
0.101-90ac87b    <  0.101.566-jb-90ac87b
0.101-a069005    <  0.101.566-jb-90ac87b
0.101-b069005    <  0.101.566-jb-90ac87b
0.101-0abc123    <  0.101.566-jb-90ac87b
```

Same base version, same code line, and the ordering flips on nothing but the abbreviated commit hash. `40690055` happens to be all digits, so the shipped 0.101 release outranks every CI build of 0.101. A hash starting with a letter loses instead, and `a`, `b` and `m` lose hardest because the comparator reads them as the alpha, beta and milestone markers.

The hash length is not stable either. `gitHash` is `ciSha.take(7)` on CI but `git rev-parse --short HEAD` otherwise, and git widens that when 7 characters would be ambiguous. The shipped `0.101-40690055` has 8 characters, so that release came from the local git path rather than the CI one.

This matters once a release and a CI build are both reachable from one IDE, which happens as soon as the plugin is on the Marketplace and also in any custom plugin repository. The update check installs whichever compares higher, so it can move a user from the release to an older CI build, or the reverse, with no relation to which was built later.

I tried a few alternative shapes against the same comparator. Only the ones that put a comparable number where the other lane puts its counter come out right:

```
[...]

Issue #270, opened 2026-07-13
Devrig autoupdate


jonnyzzz:
Parent: #294

## Release status

Safe automatic devrig update is a hard `0.103` release criterion. It depends on crash-safe installer locking/recovery in #268 and the release/compatibility contract in #295.

## Contract

- Devrig checks the explicit channel/current-release pointer from #295. It does not infer upgrade direction from semantic-version comparison, allowing an intentional rollback.
- An older running process may trigger installation but must never overwrite a launcher already switched by a newer or deliberately rolled-back operation.
- The current process is not replaced in memory; the next devrig invocation uses the activated release.
- `~/.mcp-steroid/no-autoupdate` and managed/non-interactive configuration disable automatic mutation.

## Update flow

1. Acquire or join the recoverable cross-process operation from #268.
2. Record machine-readable state under `~/.mcp-steroid/updates/`: initiating release/PID, channel, target release identity, timestamps, attempt, and status.
3. Download to a versioned staging location with bounded exponential backoff/resume.
4. Verify size, checksum, provenance/signature, and compatibility before activation.
5. Atomically switch the launcher while retaining the last known-good release.
6. Run a bounded health check; restore the known-good launcher if activation fails.
7. Publish completed/failed/rolled-back status and suppress duplicate attempts according to a documented retry policy.

## Acceptance criteria

- [ ] Multiple devrig versions/processes converge on one working target without corrupting the launcher.
- [ ] Sudden kill at every update phase leaves either the old or new launcher usable and permits recovery.
- [ ] Corrupt, missing, expired, incompatible, and partially published releases never activate.
- [ ] Status/doctor output exposes current release, channel, last check/result, active operation, known-good rollback, and opt-out reason.
- [ ] Automatic retry is bounded; failures do not create an update loop on every agent tool invocation.
- [ ] Explicit check, update, rollback, channel, and disable operations are documented for humans and automation.
- [ ] Tests cover concurrent processes, process death, network loss/resume, corrupt artifact, health failure, rollback, intentional downgrade, offline/pinned use, Windows, and Unix.
- [ ] Installer/update diagnostics never write to stdout when invoked on an MCP stdio path.

## Non-goal

Do not hot-swap the already running JVM. Update the versioned installation and launcher for subsequent processes.

jonnyzzz:
Implemented on main via #380 — ships in **v0.102**.

[...]

PR #369, opened 2026-07-29
Port VersionComparatorUtil from IntelliJ Community into mcp-core

### LINKED ISSUE #358: UpdateChecker uses startsWith instead of a version comparison
`UpdateChecker.checkForUpdates` decides whether an update exists with a string prefix test:

```kotlin
// ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/updates/UpdateChecker.kt:86
if (!currentVersion.startsWith(remoteVersion)) {
    log.info("MCP Steroid plugin update available: $remoteVersion (current: $currentVersion)")
```

`currentVersion` is the full plugin version and `remoteVersion` is `version-base` from `https://devrig.dev/version.json`. A prefix test is not an ordering, so this is wrong in both directions.

Seen on a locally built plugin under IU-263.2289:

```
Remote version: 0.101, current version: 999.0.19999-SNAPSHOT-40d32ebc
MCP Steroid plugin update available: 0.101 (current: 999.0.19999-SNAPSHOT-40d32ebc)
```

999.0 is newer than 0.101, so the notification offers a downgrade. The same happens for a real release once the base version crosses a non-prefix boundary, for example remote `0.101` against current `0.102-abc1234`.

It also misses updates. Remote `0.11` against current `0.111-abc1234` passes `startsWith`, so nothing is reported even though the versions differ.

`extractBaseVersion` already exists in the same file but is only used to format the notification text, never in the decision.

The tests do not catch this because `UpdateCheckerTest` defines its own private copies of `extractBaseVersion` and `buildUserAgent` and asserts `StringUtil.compareVersionNumbers` against those, so the branch at line 86 is never exercised. `testExtractedVersionComparison` already spells out the intended behaviour:

```kotlin
val current = extractBaseVersion("0.86.0-SNAPSHOT-20260212-193000-a1b2c3d")
assertTrue(StringUtil.compareVersionNumbers(remoteNewer, current) > 0)
```

Using that in the production path, and notifying only when the remote base is strictly greater than the extracted current base, covers all three.

To reproduce: build the plugin locally so the version becomes `<VERSION>.19999-SNAPSHOT-<hash>`, install it, and wait for the 30 second initial check.

jonnyzzz:
## Summary

Ports `VersionComparatorUtil` and its test from IntelliJ IDEA Community (Apache 2.0) into `:mcp-core`, the common module both consumers can reach: `:ij-plugin` depends on it directly, and `:npx-kt` has compile visibility via `api(project(":mcp-core"))` from `:mcp-stdio` / `:mcp-steroid-server`.

- `mcp-core/src/main/kotlin/com/jonnyzzz/mcpSteroid/util/text/VersionComparatorUtil.kt` — from `platform/util-rt/src/com/intellij/util/text/VersionComparatorUtil.java`
- `mcp-core/src/test/kotlin/com/jonnyzzz/mcpSteroid/util/text/VersionComparatorTest.kt` — from `platform/util/testSrc/com/intellij/util/text/VersionComparatorTest.java` (JUnit 3 → JUnit 5)

## Conversion notes

- Converted Java → Kotlin (`object`, Kotlin nullability instead of `org.jetbrains:annotations` — no new dependency).
- **Repackaged** to `com.jonnyzzz.mcpSteroid.util.text`: the plugin ZIP loads inside IntelliJ, which already ships the identically-named `com.intellij.util.text.VersionComparatorUtil`; a same-FQN duplicate in the plugin classloader is a hazard.
- Java `String.trim()` semantics preserved via `trim { it <= ' ' }` (Kotlin's `trim()` strips Unicode whitespace and would change tokenization).
- `max`/`min`/`COMPARATOR` keep the original's null tolerance (`String?`, `Comparator<String?>`).

## Apache 2.0 compliance

- Original JetBrains copyright headers retained; each file carries a Section 4(b) modification notice.
- New root `NOTICE` file carries IntelliJ IDEA's attribution (Section 4(d) — IntelliJ Community distributes a `NOTICE.txt`).
- NOTICE is bundled into both object-form distributions, mirroring the existing EULA wiring: plugin ZIP root, devrig dist root, and `licenses/mcp-steroid/`; pinned in both content-verification lists.

## Testing

- `./gradlew :mcp-core:test` — 4 ported tests green, no compiler warnings.
- `./gradlew :ij-plugin:verifyBundledLibraries :npx-kt:verifyBundledLibraries` — both ZIP content checks pass; NOTICE confirmed inside `mcp-steroid/NOTICE`, devrig root, and `licenses/mcp-steroid/NOTICE`.

## Review

Adversarial codex review ran on the diff (originals embedded for faithfulness checking): 1 P1 (NOTICE missing from object-form dists) + 2 P2 (trim semantics, nullability narrowing) — all fixed in the follow-up commit.

## Issues

Fixes #358 — both update checkers now gate on a real version comparison (`DevrigVersion.isUpdateAvailable`); the issue's snapshot repro (`999.0.19999-SNAPSHOT-…` offered a downgrade to `0.101`) is covered by tests, and the tests exercise the production gate instead of private copies of the logic.

Related, not closed by this PR: #360 — `DevrigVersion.comparableVersion` shields *these* update checks from the hash-vs-counter token hazard documented there, but the core problem (release and CI version-string shapes are not mutually orderable by IntelliJ's own Marketplace-side comparator) needs a version-shape change and stays open. #270 — devrig autoupdate deliberately follows a release pointer rather than version inference; this PR only fixes the notification gate.


jonnyzzz:
Round 2 (maintainer feedback): rewrote the port in idiomatic Kotlin — Kotlin's Unicode-aware `trim()` kept as a deliberate, header-documented deviation; `Regex`/`split`/`getOrNull` replace `Pattern`/`StringTokenizer`/null-padded lists; no `!!` (explicit null guard). Added the `Version` data class (`Comparable` via the comparator, so `<`/`>` operators work; BigDecimal-style equality caveat documented) and 11 new tests including this repo's real release history (0.86.0…0.101), the CI `<VERSION>.<counter>-(gh|jb)-<hash>` shape, and the `19999-SNAPSHOT` local counter sorting newest of its line.

Second adversarial codex pass: no P1 findings; semantic equivalence to the Java original verified over ~570k randomized inputs. Its three P2 advisories addressed: empty-input tokenization pinned in a direct assertion, sorted-set consequence documented on `Version`, and `_WS`/`_WORD`/`_DIGITS` names kept deliberately (the underscore prefix prevents literal token matching) with an explanatory comment.

All 19 tests green (`:mcp-core:test`), zero compiler warnings.

jonnyzzz:
Round 3: update notifications in both the plugin and devrig are now gated on a real version comparison instead of `startsWith`.

- **`DevrigVersion(value, isSnapshotBuild)`** (mcp-core): `Comparable` with the product ordering — a snapshot (local dev) build is always newer than anything promoted, so dev builds are never nagged; build metadata after the first `-` carries no precedence (semver-style). The metadata rule fixed a real bug the new tests caught: a release like `0.86.0-a1b2c3d` sorted *below* its own promoted base because the hex hash's leading `a` tokenizes as the ALPHA marker — false update nag depending on hash luck.
- **Generated metadata**: `GenerateMetadataTask` bakes `isSnapshotBuild` at build time and emits `getBuildVersion(): DevrigVersion` in both `PluginMetadata` and `DevrigVersionMetadata`; both checkers use it as the current version.
- **Shared production gate**: `DevrigVersion.isUpdateAvailable(current, promoted)` — called by `UpdateChecker` (ij-plugin) and `DevrigUpdateChecker` (devrig), and exercised directly by the tests (per codex advisory, tests no longer assert a copy of the expression).

Adversarial codex pass on the diff: no code-level P1 (the one P1 flagged the untracked new files missing from the review diff — resolved by this commit). Verified: `:mcp-core:test`, `:npx-kt:test`, full `:ij-plugin:test` (273 tests), and both distribution content verifications — all green.

PR #374, opened 2026-07-30
bin launcher: version header + never overwrite a newer launcher

### LINKED ISSUE #373: bin launcher: stamp a version header into ~/.mcp-steroid/bin/devrig and never let an older devrig overwrite a newer launcher
## Current state

`ensureBinLauncher` self-heals `~/.mcp-steroid/bin/devrig` (POSIX) / `devrig.cmd` (Windows) on every start and rewrites whenever the content differs (`BinLauncher.kt` → `writeIfChanged`). Content comparison alone means **any** devrig process — including one from an older install tree — rewrites the launcher to point at itself. With several installs present (a release, a CI build, a local dist; or two agents racing their starts), the last process to start wins regardless of which is newest.

Related to the #270 contract bullet: *"An older running process may trigger installation but must never overwrite a launcher already switched by a newer … operation."* This issue covers the launcher self-heal path only, not the full autoupdate flow.

## Proposal

1. **Version header in the generated scripts.** Machine-parseable comment carrying the generating devrig version and its source install tree:
   - POSIX: `# devrig launcher version: <full version>` + `# devrig launcher source: <install-tree launcher path>`
   - Windows: same via `rem` lines after `@echo off` (lines before `@echo off` would echo to stdout — the MCP channel).
2. **Regeneration guard.** Before rewriting, parse the existing launcher's header version and compare via `DevrigVersion` (snapshot-aware, build metadata carries no precedence): if the existing launcher's version is **strictly newer** than the running binary, skip regeneration (stderr log, no writes). A headerless or unreadable existing launcher regenerates as before (migration + corrupt-file self-heal preserved).
3. **Snapshot semantics** (deliberate, from `DevrigVersion`): a launcher written by a local SNAPSHOT build is never overwritten by a release binary — consistent with the existing SNAPSHOT default-off gate; an explicitly installed dev launcher stays pinned until a dev build rewrites it.

## Test plan

Header rendering (both OSes), header parse round-trip (`#`/`rem`/`REM`, absent header), guard truth table (older / equal / newer / snapshot-vs-release both ways / headerless), and end-to-end temp-dir tests through `ensureBinLauncherCore` (newer existing file left byte-identical; older existing file rewritten).

jonnyzzz:
Fixes #373.

## What

- The generated `~/.mcp-steroid/bin/devrig` (POSIX) / `devrig.cmd` (Windows) launchers now carry a machine-parseable header: `# devrig launcher version: <version>` + `# devrig launcher source: <install-tree launcher>` (`rem` equivalents after `@echo off` on Windows).
- Before regenerating, the self-heal parses the existing launcher's header and **skips the write when its version is strictly newer** than the running binary — an older devrig never re-registers itself over a newer launcher. Comparison is `DevrigVersion` (snapshot always outranks a release; build metadata after the first `-` carries no precedence). Headerless (pre-#373) or unreadable launchers regenerate as before.

## Design notes

- **No cross-process lock.** Launcher replacement is an atomic same-directory move (`writeAtomically`: temp file + `ATOMIC_MOVE`), so concurrent starts can never produce a torn file. The worst case without serialization is a stale-read guard decision by a racing process, which the newer devrig self-heals on its next start — not worth a lock on the latency-sensitive `devrig mcp` start path.
- A kept newer launcher still gets a lost executable bit repaired (otherwise no devrig would ever fix it).
- The cmd header **quotes** the source path — `rem` does not neutralize `&`, so an unquoted install path like `C:\tools & extras\…` would execute the tail; header interpolations strip CR/LF so a value cannot escape its comment line.

## Testing

`:npx-kt:test` green — 21 BinLauncherTest cases including: header rendering on both OSes, parse round-trip (`#`/`rem`/`REM`, headerless → null), guard truth table (older/equal/newer + both snapshot directions), end-to-end older-never-overwrites / newer-regenerates temp-dir flows, metacharacter + newline sanitization, and +x repair on a kept launcher.

### DIFF
--- a/npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/BinLauncher.kt
+++ b/npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/BinLauncher.kt
@@ -1,6 +1,7 @@
 /* Copyright 2025-2026 Eugene Petrenko (mcp@jonnyzzz.com); Copyright 2025-2026 JetBrains. Use of this source code is governed by the Apache 2.0 license. */
 package com.jonnyzzz.mcpSteroid.devrig
 
+import com.jonnyzzz.mcpSteroid.util.text.DevrigVersion
 import java.io.File
 import java.nio.file.Files
 import java.nio.file.Path
@@ -82,7 +83,9 @@ private fun parseBinAutoRegisterOptOut(value: String?): Boolean? = when (value?.
  * The launcher is rewritten ONLY when its content actually changed (normalized comparison), never on
  * every launch — and writes are atomic (temp file + atomic move), so a concurrent agent that is mid-read
  * of the file (the contract: it can change WHILE the binary runs) never sees a torn launcher. When the
- * content already matches, a lost executable bit is still repaired.
+ * content already matches, a lost executable bit is still repaired. An existing launcher whose stamped
+ * header version is strictly newer than this binary is never overwritten at all — see
+ * [shouldKeepNewerLauncher] (#373).
  *
  * [force] = an explicit `devrig install`: write regardless of the SNAPSHOT/dev passive-start default
  * (but still honoring an explicit opt-out), so install never registers a wrapper it didn't write.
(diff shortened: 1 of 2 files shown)


jonnyzzz:
## Review 2026-08-03

- **Issue #373 is already CLOSED by owner design decision**, not by this PR: the exact mechanism here (launcher version stamp + no-downgrade guard) existed in an early revision of the #380 auto-update work and was **deliberately removed** during the simplification rounds under the principle "the devrig binary is not a source of truth" (see the close comment on #373).
- Main's spec now codifies the prohibition: `docs/updates-check/devrig-auto-update.md` (Design decisions, ~line 249) — *"No launcher version stamps, no content verification, no completion markers"*; launcher correctness is owned by the rename-based replacement instead. The flip-back window this PR guards against is an explicitly accepted Tradeoff 3 (self-corrects on release cadence).
- The concrete overwrite hazard was fixed at the root by #385 (commit 64929ec8, rename-based `.new<pid>` → move → `.old<pid>` sequence), and 6c5db166 explicitly *dropped* locks and stamps — this PR's sibling-file `tryLock` in `BinLauncher.kt` re-introduces machinery the merged design removed. Multi-install last-start-wins on dev boxes stays intentional (SNAPSHOT gate covers it).
- Textually stale: `mergeable=CONFLICTING` / `mergeStateStatus=DIRTY` — main rewrote the `BinLauncher.kt` write path after this PR's single commit (64929ec8, 6c5db166, 11c0a7f0/#410).
- The implementation itself is solid (guard truth table incl. both snapshot directions, cmd `&`-quoting with quoted source path, CR/LF header sanitization, lock-skip on the hot path) — no banned-pattern violations. If launcher-version pinning is ever wanted again, #373's close note points at PR #380's pre-simplification history as the recoverable source; this branch is a second one.

**Verdict: superseded** — the #380/#385 auto-update series resolved #373 by design decision on main; merging this would reverse a documented, quorum-reviewed owner call rather than fix a gap. Recommend closing.

jonnyzzz:
Simplified per review: dropped the sibling-file `tryLock` around the self-heal (commit b93a87f4). Replacement is an atomic same-directory move (`writeAtomically`: temp file + `ATOMIC_MOVE`), so the race the lock serialized has a benign worst case — a stale-read guard decision, self-healed on the newer devrig's next start, never a torn file. The #373 guard (an older devrig never re-registers over a newer launcher) is unchanged. Lock test removed; `:npx-kt:test` green (21 BinLauncherTest cases). Ready for re-review.

jonnyzzz:

PR #385, opened 2026-07-30
Launcher replacement is rename-based to survive Windows file locks

jonnyzzz:
## Problem

`ensureBinLauncher` self-heals `~/.mcp-steroid/bin/devrig` (POSIX sh) / `devrig.cmd` (Windows batch) on every devrig start and from `devrig install devrig`, via a temp file + `Files.move(tmp, target, ATOMIC_MOVE, REPLACE_EXISTING)`. On Windows this fails whenever another process holds the launcher open — most notably **cmd.exe executing `devrig.cmd` right now** (the wrapper that launched the very devrig doing the rewrite), or an antivirus/indexer scanning it. The self-heal then silently no-ops (best-effort catch), leaving a stale launcher in place.

## Research: why replace fails and rename works

**What Java does on Windows.** OpenJDK's [`WindowsFileCopy.move`](https://github.com/openjdk/jdk/blob/master/src/java.base/windows/classes/sun/nio/fs/WindowsFileCopy.java) implements `ATOMIC_MOVE` as `MoveFileEx(src, dst, MOVEFILE_REPLACE_EXISTING)`; the non-atomic `REPLACE_EXISTING` path is `DeleteFile(target)` followed by `MoveFileEx(..., 0)`. Both need delete-class access to the **target**. Per Microsoft's [Moving and Replacing Files](https://learn.microsoft.com/en-us/windows/win32/fileio/moving-and-replacing-files) and the [MoveFileEx docs](https://learn.microsoft.com/en-us/windows/desktop/api/WinBase/nf-winbase-movefileexa), that fails when the destination is open without compatible sharing — surfacing as `ERROR_SHARING_VIOLATION` (→ `FileSystemException`) or `ERROR_ACCESS_DENIED` (→ `AccessDeniedException`).

**Why a rename-aside succeeds.** A rename rewires only the directory entry, never the file data. Renaming a file another process has open succeeds when the open handles share delete access (`FILE_SHARE_DELETE`), and — the key asymmetry — a running or memory-mapped file blocks *delete and write* but **not rename**: [Len Holgate, "Allowing renaming but not deleting for files on Windows"](https://lenholgate.com/blog/2022/09/allowing-renaming-but-not-deleting-for-files-on-windows.html); the Go toolchain relies on the same behavior ([golang/go#21997](https://github.com/golang/go/issues/21997): the executable of a running process can be renamed without error — rename only, not delete or write). Handles follow the renamed file, so whoever holds the old launcher keeps reading exactly the bytes it opened.

**The cmd.exe batch hazard.** cmd.exe reads batch files incrementally, remembering a byte offset between lines ([SS64](https://ss64.com/nt/syntax-redirection.html), [The Old New Thing on batch parsing quirks](https://devblogs.microsoft.com/oldnewthing/20210726-00/?p=105483)) — the classic reason editing a running batch file misbehaves: the interpreter resumes at the old offset inside *new* bytes. An in-place overwrite of `devrig.cmd` while cmd executes it risks exactly that; a rename leaves the old bytes intact under cmd's position, and the new file only takes over the *name*.

**Prior art.** This is the standard self-replacement dance on Windows: [rustup's self-update](https://github.com/rust-lang/rustup/blob/main/src/cli/self_update.rs) parks the live binary as `.old` and moves the new one in ([overview](https://deepwiki.com/rust-lang/rustup/4-self-update-system)); [`self-replace`](https://docs.rs/self-replace) and [go-update](https://github.com/inconshreveable/go-update/issues/14) do the same.

**ATOMIC_MOVE applicability.** All renames here are same-directory, same-volume, so `MoveFileEx` handles them as true renames; `AtomicMoveNotSupportedException` only occurs on `ERROR_NOT_SAME_DEVICE`, which cannot happen here — the sequence's non-atomic second attempt covers exotic filesystems anyway.

## Mechanism: one fixed sequence

`replaceLauncherFile(target, content, executable)` in `BinLauncher.kt` — identical on every platform, identical whether or not the original file exists, no existence checks:

1. write the new launcher content to `<name>.new<pid>` (executable bit set before any move) — the original is never edited in place;
2. attempt to move it onto the original **atomically**;
3. attempt to move it **non-atomically**;
4. rename the original to `<name>.old<pid>` — the rename succeeds where the replace was blocked (see research);
5. repeat the two move attempts;
6. delete `<name>.old<pid>`.

If anything fails at any step, wait 10 ms and repeat the whole sequence — up to **5 attempts total** — then log to stderr and give up (the next devrig start rewrites the launcher anyway). A no-op — nothing to park, nothing to delete — is success, not failure, so a missing original converges on the first move.

**Accepted trade-offs (by design):** the brief no-launcher gap between step 4 and step 5; a tiny `.new<pid>`/`.old<pid>` leftover after a crash or after 5 failed rounds (each process touches only its own pid-named files — there is no cross-process sweeping or directory scanning).

## Tests (`BinLauncherTest`, all green via `./gradlew :npx-kt:test`)

The Windows lock scenario is exercised on POSIX through the `move` seam, which every rename in the sequence goes through:

- a seam that fails moves onto an *existing* target (modeling the sharing violation) but allows renames → the original is parked as `devrig.old<pid>` (verified with its original bytes mid-sequence), the target gets the new content with the executable bit intact, and the `.old<pid>` is deleted as the last step (4 seam calls total)
- a missing original converges within the first attempt: one atomic move, nothing parked, no leftovers
- when every move keeps failing: 5 rounds (15 seam calls) with 10 ms delays in between (asserted ≥ 40 ms elapsed), then the last failure propagates loudly and the original launcher is left untouched

## Validated on real Windows (eugene-x220)

Windows 10 22H2 (19045) / NTFS / PowerShell 5.1 / Corretto JDK. A dedicated **`BinLauncherWindowsTest`** (`@EnabledOnOs(WINDOWS)`) exercises the REAL sequence — no seam injection — against genuine holder processes. Results on the box: **`BinLauncherWindowsTest` 4/4 passed**, and the full **`:npx-kt:test` suite: 1315 tests, 0 failures, 0 errors, 25 skipped** (first time this module ever ran on real Windows).

| Scenario | Holder | Observed on NTFS |
|---|---|---|
| No contention | — | create + replace both work, no leftovers (0.01 s) |
| **cmd.exe EXECUTING the launcher** | `cmd /d /c` running a looping batch | **the direct atomic replace succeeded on the very first move** — cmd.exe opens executing batches with full sharing (read+write+delete), so the production case (rewriting `devrig.cmd` from the devrig it launched) does not even need the fallback; stderr empty, zero leftovers |
| **Memory-mapped section** (running-file semantics) | PowerShell `MemoryMappedFile.CreateFromFile` over a delete-sharing handle | atomic and plain replace onto the mapped file → `AccessDeniedException`; **the rename-aside succeeded** (`devrig.cmd` parked as `devrig.cmd.old<pid>`), the new content landed; the aside delete failed on all 5 rounds while the mapping lived (`AccessDeniedException: ...devrig.cmd.old8704`) → loud give-up with the CORRECT content in place; the aside was deletable the moment the holder died |
| **`FileShare.Read`-only** (strict scanner) | PowerShell `File.Open` | everything blocked *including* the rename-aside ("The process cannot access the file because it is being used by another process") — confirming a rename needs delete sharing; 5 rounds, original launcher untouched and usable; the first write after the holder exited succeeded |

The three outcomes match the researched semantics exactly: full-sharing holders (cmd.exe) never block; mapped/running files block replace but not rename — the case this PR fixes; no-delete-sharing holders block everything and degrade to the designed give-up + next-start self-heal.

### DIFF
--- a/npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/BinLauncher.kt
+++ b/npx-kt/src/main/kotlin/com/jonnyzzz/mcpSteroid/devrig/BinLauncher.kt
@@ -2,7 +2,6 @@
 package com.jonnyzzz.mcpSteroid.devrig
 
 import java.io.File
-import java.nio.file.AtomicMoveNotSupportedException
 import java.nio.file.Files
 import java.nio.file.Path
 import java.nio.file.StandardCopyOption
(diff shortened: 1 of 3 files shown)


jonnyzzz:
Cross-reference from the auto-update design (quorum nit): the auto-updater supervises the installer with a 30-min timeout that force-kills the whole process tree — the kill can land mid-`devrig install devrig`, i.e. mid-launcher-replacement. Please double-check the fallback's kill-window: between the aside-rename and the into-place rename there is a moment with no `bin/devrig` on disk, and the current rollback handling covers a failed second move but not a SIGKILL between the two. Agents spawn devrig by that absolute path, and a missing launcher has no self-heal trigger (nothing running can rewrite it if no devrig can start). Worth either swapping the order into a strictly crash-safe sequence (e.g. hardlink/copy-first) or documenting the window as negligible (two syscalls) with the manual-reinstall recovery path.

jonnyzzz:
Re the kill-window concern (auto-update supervisor force-killing mid-replacement): the algorithm was reworked per the owner's direction, and the gap between the aside-rename and the into-place rename is now **explicitly accepted** ("we need to be as atomic as possible... The small gap with the file availability is acceptable"). No extra machinery (hardlink/copy-first, rollback) is added to close it, by design.

What covers the two failure modes instead:

- **Transient failure / racer:** any step failing retries the whole remaining sequence from the current state after 10 ms, up to 5 attempts total. A racer hitting the two-rename gap (two same-directory renames, microseconds apart) is caught by its own retry pass.
- **SIGKILL between the renames:** the on-disk state is then `devrig.old<pid>` (the old, fully working launcher) + `devrig.new<pid>` (the staged new one) and no `devrig`. Recovery is the next devrig start's self-heal: agents that spawn devrig by the *install-tree* path (or `devrig install devrig` re-run from the install script) rewrite the launcher unconditionally, and the start-time sweep now also reaps both `old<pid>` asides and crash-orphaned `new<pid>` staging files. The previous rollback-the-aside logic was dropped as agreed — the retry pass plus next-start self-heal cover it.

The window itself is two same-volume `MoveFileEx` renames back to back, and it is only entered at all when the direct atomic replace (step 2) is blocked by a lock — on POSIX it never happens.

jonnyzzz:

PR #397, opened 2026-07-31
0.102 reliability: devrig mcp orphan watchdog, friendly missing-CLI errors, EDT-safe VFS refresh

### LINKED ISSUE #132: devrig mcp: orphaned stdio processes leak (no parent-death watchdog)
## Symptom

Multiple `devrig mcp` (stdio MCP server) processes accumulate over time and never exit. Observed **10 live `devrig mcp` processes** on one machine — ages from minutes to **4 days** — at ~250 MB RSS each (≈ **2.5 GB** wasted). They sit idle (0% CPU), but the pile-up wastes RAM, runs several devrig instances that all discover/bridge the same IDE, and each fresh spawn's JIT warmup briefly spikes CPU (looked like "stuck MCP Steroid").

## Root cause

`devrig mcp`'s read loop (`McpStdioServer`) exits cleanly **on stdin EOF**:

```kotlin
while (currentCoroutineContext().isActive) {
    val n = readChunk(buf)   // input.read(buf) on Dispatchers.IO
    if (n < 0) break         // EOF
    ...
}
```

So a normally-closed client (the parent closes the stdin pipe) reaps fine. But an **orphaned** `devrig mcp` — whose parent process died **without** closing the stdin pipe (e.g. a `SIGKILL`'d Claude/Codex session) — **never receives EOF** and lives forever. There is **no parent-liveness watchdog**, so orphans accumulate across sessions.

Verified: the leaked processes all had a (still-listed) parent pid but came from long-abandoned `claude`/`codex` sessions; killing the devrig children with `SIGTERM` freed the memory cleanly.

## Proposed fix

Add a parent-death watchdog to `devrig mcp` that exits the process when its launching parent is gone, complementing the EOF path so orphans self-terminate even when stdin never closes.

### Cross-platform (must work on macOS, Linux, AND Windows)

- **Portable (preferred):** record the launching parent pid at startup, then poll `java.lang.ProcessHandle.of(parentPid).map { it.isAlive }.orElse(false)` on a small daemon timer; exit when it goes false. `ProcessHandle` is JDK 9+ and works on Windows, macOS, and Linux — no native code, one implementation.
- **POSIX-only signals (insufficient alone):** `getppid() == 1` (reparented to init/launchd) is Unix-only — **does not exist on Windows** (no reparenting-to-init, no `getppid`), so it cannot be the primary mechanism. At most a fast-path supplement on Unix.
- **Windows-specific alternative:** assign the child to a **Job Object** with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE` so the OS terminates it when the parent (job owner) dies. Robust but parent-side + JNA/native — heavier than the portable `ProcessHandle` poll.
- **stdin EOF semantics also differ across platforms** (Windows anonymous-pipe handle inheritance/closure vs Unix pipes), which is another reason not to rely on EOF alone; the `ProcessHandle`-poll watchdog covers every OS uniformly.

## Repro
Start `devrig mcp` from a parent, kill the parent so stdin isn't closed (`SIGKILL` on Unix; `taskkill /F` on Windows), observe `devrig mcp` lingering indefinitely.

_Found 2026-06-20 while debugging an apparently-stuck MCP Steroid connection._

### LINKED ISSUE #318: syncDocuments runs recursive VFS refresh on EDT and freezes the IDE for 30s
## Summary

`steroid_execute_code` with the default `smart_non_modal` profile can freeze the IntelliJ UI for 30+ seconds and trigger the IDE warning that MCP Steroid may be slowing it down.

The freeze is real and correctly attributed to the plugin: `McpScriptContextImpl.syncDocuments()` invokes a recursive project-root VFS refresh while running inside `Dispatchers.EDT`.

## Observed failure

Environment:

- IntelliJ IDEA 2026.1, build `IU-261.25134.95`
- MCP Steroid direct IDE connection
- A `smart_non_modal` `steroid_execute_code` call polling a test run after the preceding call had generated build output

Relevant `idea.log` sequence:

```text
16:47:03,092 ScriptExecutor - Running ... modal=SMART_NON_MODAL
16:47:34,349 VfsRefreshService - awaitRefresh timed out after 30 s
16:47:34,359 PerformanceWatcherImpl - UI was frozen for 31255ms
  ...threadDumps-freeze-...-RefreshWorker.processQueue-31sec
```

The generated freeze report attributes the EDT stack to MCP Steroid:

```text
java.awt.EventDispatchThread.run
  ...
  com.jonnyzzz.mcpSteroid.execution.McpScriptContextImpl$syncDocuments$2$1.invokeSuspend
    com.jonnyzzz.mcpSteroid.execution.VfsRefreshService.awaitRefresh
      com.intellij.openapi.vfs.newvfs.RefreshQueueImpl.refresh
        com.intellij.openapi.vfs.newvfs.RefreshWorker.processQueue
          com.intellij.openapi.vfs.newvfs.RefreshWorker.fullDirRefresh
            com.intellij.openapi.vfs.impl.local.LocalFileSystemImpl.listWithAttributes
              java.nio.file.Files.isWritable
                sun.nio.fs.UnixNativeDispatcher.access0
```

The scan was inside a non-cancellable section and native filesystem calls, so the 30-second coroutine timeout did not keep the UI responsive. It only logged after the blocking traversal returned.

## Root cause

`McpScriptContextImpl.syncDocuments()` puts commit, save, **and the recursive refresh** into one EDT block:

```kotlin
withContext(Dispatchers.EDT) {
    PsiDocumentManager.getInstance(project).commitAllDocuments()
    FileDocumentManager.getInstance().saveAllDocuments()
    project.vfsRefreshService.awaitRefresh()
}
```

`VfsRefreshService.awaitRefresh()` then recursively refreshes the complete project root:

```kotlin
withTimeoutOrNull(30.seconds) {
    RefreshQueue.getInstance().refresh(recursive = true, listOf(base))
}
```

Although its KDoc says that the coroutine-native overload runs as a background write action, the captured runtime stack shows the actual directory scan executing on `AWT-EventQueue-0` when called from this EDT context.

This work is also duplicated:

- `CodeEvalManager` already calls `awaitRefresh()` before script compilation from its background execution path.
- The default `smart_non_modal` profile calls `syncDocuments()` before and after the script.
- `ExecutionManager` also schedules the asynchronous tail refresh.

A build/test producing many files makes the redundant recursive pre-flight scan particularly expensive.

## Test gap

`VfsRefreshServiceTest` explicitly uses:

```kotlin
override fun runInDispatchThread(): Boolean = false
```

and only verifies that `awaitRefresh()` completes within 35 seconds. It therefore cannot detect that a 30-second traversal monopolizes the EDT.

## Expected behavior

- No recursive filesystem traversal runs on EDT.
- `smart_non_modal` keeps its document/VFS consistency contract without making the IDE unresponsive.
- A slow refresh may delay the MCP response, but it must not freeze the UI or trigger the plugin-performance warning.
- The timeout remains cancellation-safe and diagnostic.

## Suggested direction

Keep only the operations that require EDT in the EDT block, and run any awaited refresh after leaving it:

```kotlin
withContext(Dispatchers.EDT) {
    PsiDocumentManager.getInstance(project).commitAllDocuments()
    FileDocumentManager.getInstance().saveAllDocuments()
}
project.vfsRefreshService.awaitRefresh()
```

Alternatively, determine whether the awaited refresh should be removed from `syncDocuments()` entirely, given the existing background pre-compilation refresh and asynchronous post-execution refresh. That requires preserving the documented stale-VFS guarantees.

## Acceptance criteria

- Add a regression test proving `syncDocuments()` never invokes the recursive refresh on EDT.
- Exercise a deliberately slow/delayed refresh while verifying EDT remains responsive.
- Preserve commit/save behavior and the next-call VFS consistency contract.
- Verify a normal `smart_non_modal` call no longer produces a `PerformanceWatcher` freeze attributed to MCP Steroid.

### LINKED ISSUE #342: devrig install <agent> --check crashes with a raw stacktrace when the agent CLI is missing (and: is --check needed given install is idempotent?)
### Observed
In a clean environment (no `claude` on PATH), running `devrig install claude --check` fails with a bare Java stacktrace:

```
java.io.IOException: Cannot run program "claude": error=2, No such file or directory
```

and exits with code `64`. Reproduced in a fresh tart macOS VM with devrig 0.101.

### Expected
A graceful, human-readable message like:

```
claude not found on PATH — install it first
```

(a non-zero exit is fine), NOT an unhandled stacktrace.

### Additional design question
Since `devrig install <agent>` is already **idempotent** (it upserts the MCP registration — remove old + add canonical each run), is a separate `--check` mode needed at all? Consider whether `--check` should be **removed/folded into** the idempotent install rather than fixed, to keep a single code path.

### Scope
Small CLI UX/robustness issue found during devrig.dev site QA. No security impact.

jonnyzzz:
Three 0.102 reliability fixes from the "try the remaining 0.102 issues" round. Test-first; each commit atomic.

Fixes #132, fixes #342, fixes #318.

Scope note: a fourth fix (#179, inspection-sweep cancellation) was built, dual-reviewed, then **parked and fully reverted by owner decision** — and subsequently re-validated as **obsolete on IU-2026.1**: live probes show `ProgressManager.checkCanceled()` now observes job cancellation even under an installed `EmptyProgressIndicator` (details + probe data on the closed issue).

## What changed

| Commit | Issue | Change |
|---|---|---|
| 7cfa00cc | #132 | `ParentDeathWatchdog` for `devrig mcp`: ProcessHandle liveness poll (5s × 2 confirmations) on the launching ancestors; orphans self-terminate when the agent dies without closing stdin. Unknown ancestry → disabled (never false-positive) |
| 9dbd4397 | #342 | `devrig install <agent>` / `--check` with a missing CLI prints guidance + exit 64 instead of a raw IOException stacktrace |
| 2bec3232 | #318 | `syncDocuments()` awaits the recursive VFS refresh OFF the EDT (commit/save stay on it); `awaitRefresh` fails fast if ever awaited from the EDT again (the captured 31s RefreshWorker-on-EDT freeze) |
| e8106829 | — | Review fixes: Windows `cmd.exe`-wrapper grandparent watching for #132 (the .cmd launcher can't exec, so `taskkill /F` on the agent leaves the wrapper alive); typed `AgentCliNotLaunchableException` so infrastructure IOExceptions are not misreported as a missing CLI |

## Verification

- `:npx-kt:test`, `:ai-agents:test` green; `:ij-plugin:test` full suite green (282 tests) on the pre-revert tree, scoped re-run green after the #179 revert.
- Dual-model adversarial review (Fable-5 + gpt-5.6-sol) over the pre-revert range:
  - gpt-5.6-sol: 2 findings on the surviving fixes, both applied (#132 Windows-wrapper grandparent watching — real-Windows verification still pending on the issue; #342 typed launch exception). Its remaining findings targeted the reverted #179 work.
  - Fable-5: explicitly no findings on #132/#342/#318 across all lenses (ProcessHandle semantics, IOException scoping, VFS/read-action rules, CLAUDE.md conventions). Its single BLOCKER targeted the reverted #179 work and was empirically refuted before the revert.
- Windows note: the #132 grandparent-watch logic is unit-tested; a real `taskkill /F` process-tree verification on eugene-x220 is tracked on the issue.

### DIFF
--- a/ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/execution/VfsRefreshService.kt
+++ b/ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/execution/VfsRefreshService.kt
@@ -99,9 +99,12 @@ class VfsRefreshService(
         // PerformanceWatcher freeze attributed to this plugin was captured with the RefreshWorker
         // traversal on AWT-EventQueue-0. Fail fast so no caller can reintroduce that: every await
         // must happen on a background dispatcher (syncDocuments awaits AFTER its EDT commit/save
-        // block; CodeEvalManager's pre-compile path is background already).
-        check(!ApplicationManager.getApplication().isDispatchThread) {
-            "awaitRefresh must not be awaited from the EDT — the recursive VFS scan would freeze the UI (#318)"
+        // block; CodeEvalManager's pre-compile path is background already). Plain if+throw on
+        // purpose — this guard is unconditional, fixed behavior, never elidable.
+        if (ApplicationManager.getApplication().isDispatchThread) {
+            throw IllegalStateException(
+                "awaitRefresh must not be awaited from the EDT — the recursive VFS scan would freeze the UI (#318)",
+            )
         }
         val base = projectBaseVf() ?: return
         // Use the platform's coroutine-native [RefreshQueue.refresh] (suspend
--- a/ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/execution/VfsRefreshService.kt
+++ b/ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/execution/VfsRefreshService.kt
@@ -1,6 +1,7 @@
 /* Copyright 2025-2026 Eugene Petrenko (mcp@jonnyzzz.com); Copyright 2025-2026 JetBrains. Use of this source code is governed by the Apache 2.0 license. */
 package com.jonnyzzz.mcpSteroid.execution
 
+import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.components.Service
 import com.intellij.openapi.components.service
 import com.intellij.openapi.diagnostic.thisLogger
@@ -94,6 +95,14 @@ class VfsRefreshService(
      * pathological refresh cannot hang compilation forever.
      */
     suspend fun awaitRefresh() {
+        // #318: a recursive project-root scan awaited from the EDT monopolizes it — a 31s
+        // PerformanceWatcher freeze attributed to this plugin was captured with the RefreshWorker
+        // traversal on AWT-EventQueue-0. Fail fast so no caller can reintroduce that: every await
+        // must happen on a background dispatcher (syncDocuments awaits AFTER its EDT commit/save
+        // block; CodeEvalManager's pre-compile path is background already).
+        check(!ApplicationManager.getApplication().isDispatchThread) {
+            "awaitRefresh must not be awaited from the EDT — the recursive VFS scan would freeze the UI (#318)"
+        }
         val base = projectBaseVf() ?: return
         // Use the platform's coroutine-native [RefreshQueue.refresh] (suspend
         // overload). The callback-based variant uses an EDT-side
--- a/ai-agents/src/main/kotlin/com/jonnyzzz/mcpSteroid/aiAgents/AiAgentCli.kt
+++ b/ai-agents/src/main/kotlin/com/jonnyzzz/mcpSteroid/aiAgents/AiAgentCli.kt
@@ -1,6 +1,7 @@
 /* Copyright 2025-2026 Eugene Petrenko (mcp@jonnyzzz.com); Copyright 2025-2026 JetBrains. Use of this source code is governed by the Apache 2.0 license. */
 package com.jonnyzzz.mcpSteroid.aiAgents
 
+import java.io.IOException
 import java.nio.file.Files
 import java.util.concurrent.TimeUnit
 import kotlin.time.Duration
@@ -53,6 +54,18 @@ fun interface AiAgentCliRunner {
 }
 
 /**
+ * The agent CLI process could not be LAUNCHED — the binary is missing from PATH, or is a shape the
+ * OS cannot spawn directly (a Windows `.cmd` npm shim needs `cmd.exe /d /c`). Distinct from other
+ * IOExceptions (temp-file creation, output read) so callers can turn exactly this case into
+ * "install the CLI first" guidance (jonnyzzz/mcp-steroid#342) without masking infrastructure
+ * failures as a missing CLI.
+ */
+class AgentCliNotLaunchableException(
+    val binary: String,
+    cause: IOException,
+) : IOException("could not launch '$binary': ${cause.message}", cause)
+
+/**
  * Runs an agent CLI to completion with a hard [timeout].
  *
  * Output goes to a TEMP FILE, not a pipe: with a pipe, draining via
@@ -73,10 +86,14 @@ class ProcessAiAgentCliRunner(
     override fun run(invocation: AiAgentCliInvocation): AiAgentCliResult {
         val outputFile = Files.createTempFile("devrig-agent-cli-", ".out")
         try {
-            val process = ProcessBuilder(listOf(invocation.binary) + invocation.args)
-                .redirectErrorStream(true)
-                .redirectOutput(outputFile.toFile())
-                .start()
+            val process = try {
+                ProcessBuilder(listOf(invocation.binary) + invocation.args)
+                    .redirectErrorStream(true)
+                    .redirectOutput(outputFile.toFile())
+                    .start()
+            } catch (e: IOException) {
+                throw AgentCliNotLaunchableException(invocation.binary, e)
+            }
             runCatching { process.outputStream.close() } // stdin: immediate EOF
             if (!process.waitFor(timeout.inWholeMilliseconds, TimeUnit.MILLISECONDS)) {
                 process.destroyForcibly()
--- a/ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/execution/McpScriptContextImpl.kt
+++ b/ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/execution/McpScriptContextImpl.kt
@@ -307,8 +307,14 @@ class McpScriptContextImpl(
                 withContext(Dispatchers.EDT) {
                     PsiDocumentManager.getInstance(project).commitAllDocuments()
                     FileDocumentManager.getInstance().saveAllDocuments()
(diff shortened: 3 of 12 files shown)


ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/execution/VfsRefreshService.kt
```diff
@@ -94,6 +95,14 @@ class VfsRefreshService(
      * pathological refresh cannot hang compilation forever.
      */
     suspend fun awaitRefresh() {
+        // #318: a recursive project-root scan awaited from the EDT monopolizes it — a 31s
+        // PerformanceWatcher freeze attributed to this plugin was captured with the RefreshWorker
+        // traversal on AWT-EventQueue-0. Fail fast so no caller can reintroduce that: every await
+        // must happen on a background dispatcher (syncDocuments awaits AFTER its EDT commit/save
+        // block; CodeEvalManager's pre-compile path is background already).
+        check(!ApplicationManager.getApplication().isDispatchThread) {
```
jonnyzzz: should be just if -- throw error and give up, it should not be turned off like check function is controlled with -ea for the JVM -- we need fixed behaviour

installer-gen/src/main/resources/templates/install.ps1.tmpl
jonnyzzz: done in #399

jonnyzzz:
Review comments processed:

- **VfsRefreshService EDT guard** → done in 4fba3dfd: explicit `if + throw IllegalStateException`, unconditional fixed behavior. (For the record, Kotlin's `check()` always throws and is not `-ea`-gated like Java's `assert` — but the explicit form leaves no room for doubt.)
- **install.ps1.tmpl session-PATH / guidance** → resolved by the rebase onto #399's main: this branch no longer touches the installer templates at all; the guidance now lives in `devrig install devrig` per the locked install-scripts contract (docs/install-scripts-contract.md).

Note: there is still an unsubmitted pending review on this PR from your account (it blocked in-thread replies) — submit or discard it when convenient.

jonnyzzz:
All four pending-review comments processed in 67e3b631:

- **install.sh.tmpl "done in #399"** → moot after the rebase: this branch no longer touches either installer template.
- **HelpCommand "merges to `devrig install [claude|codex|gemini|...]`"** → the bare and agent-qualified forms now share ONE help entry: `devrig install [claude|codex|gemini] [--check]` (with-no-agent behavior documented inside it); DevrigCommandOutputTest updated.
- **InstallCommand "check if we already have that logic somewhere up the stack"** → yes, the pattern exists one level up: runCli's typed-exception catches (ManagedBackend*Exception → friendly stderr + 64). The AgentCliNotLaunchableException handling moved there (`reportAgentCliNotLaunchable`, unit-tested); the inline wrappers are gone and the pure install/check functions propagate the typed exception untouched (pinned by test).
- **ParentDeathWatchdog "no need for companion object"** → `log` is a plain instance field now.

`:npx-kt:test` green (1370+ tests). The pending review itself is still unsubmitted on your side.

Issue #423, opened 2026-08-02
`devrig backend download idea-community` always fails — baseline build compared to full build, after an 850 MB download


jonnyzzz:
**Severity: highest of this set.** The first IDE an agent picks for a Java project is Community, and
that path cannot succeed on any current build.

**Problem.** `GithubCommunityReleases.kt:115` sets `build = baseline.toString()` — the file's own
comment says so: *"GitHub releases do not expose the full build number, but the baseline (which is
[what we can derive])"*. So a resolved `idea-community 2026.2` carries `build = "262"`. After the
download, `validateInstalledBuildNumber` (`ManagedBackend.kt:513`) compares that against the real
`product-info.json` build via `ideBuildMatchesProduct` (`ManagedBackend.kt:1296`), which is exact
string equality after stripping a `IC-`/`IU-` prefix. `"262" != "262.8665.258"` → always throws:

```
Downloaded idea-community 2026.2 resolved to build 262, but product-info.json under
/home/agent/.mcp-steroid/backends/idea-community-2026.2.partial/5TDxQOPLoOghxb6i-idea-oss
reports 262.8665.258. The install is not the requested IDE artifact.
```

IU is unaffected because the JetBrains download feed returns the exact build (`262.8665.337`), which
is why PR #411 never hit this.

The failure lands **after** the full ~850 MB download, and it is deterministic — retrying downloads
the archive again and fails identically.

**Reproduction** (linux/aarch64, devrig `0.101.19999-SNAPSHOT-0cf1b655`) — hit independently by both
probe agents, each of which picked Community first as the obvious choice for a Java/Maven project:
```bash
devrig backend download idea-community        # ~850 MB, then ManagedBackendValidationException
devrig backend download idea-community        # same failure, downloads again
devrig backend download idea-community --json # exit 64
```

**Proposal.**
- When the resolver only knows a baseline, validate as a **baseline prefix**: accept `actual` whose
  first dot-separated segment equals `expected` (`262.8665.258` matches `262`). Keep exact matching
  when the resolver produced a full build.
- Mark the resolution as baseline-only (e.g. `BackendDownloadResolution.buildIsBaseline`) so the
  check picks the right comparison instead of inferring from the string shape.
- Validate the build number **before** committing to the download where possible, or at minimum keep
  the unpacked archive so a retry does not re-download 850 MB.

---

[...]

PR #431, opened 2026-08-02
devrig: integration-test the IDE download path for every supported product (incl. Android Studio)

jonnyzzz:
#429 fixed the baseline-vs-full build comparison that rejected every `idea-community` download, but
the regression test that came with it only covered `idea-community`. The same defect lived in
`AndroidStudioCanaryReleases` — the canary resolver derives the build the same way — and nothing
asserted the contract for the other seven products at all.

This adds coverage for **all nine supported IDEs across all three feeds**, Android Studio included.

## What is tested

`AllIdeProductsDownloadTest` (default `:npx-kt:test`, offline) walks every `IdeProduct.knownProducts`
entry through the production path: feed dispatch → the real per-feed resolver over a recorded payload
→ `BackendManager.download` → unpack → `product-info.json` validation → descriptor. Per product it
asserts

- the feed's build shape and `buildIsBaseline` agree (only `data.services.jetbrains.com` publishes a
  full build; GitHub Community and the Android Studio preview page stop at the platform baseline),
- the resolved build is loadable by the bundled plugin (`plugin.xml` `since-build 261`),
- the build a real install of that artifact reports **passes** validation (`262` ↔ `262.8665.258`,
  `261` ↔ `AI-261.24374.151.2611.14049965`),
- an artifact from the next baseline is still **refused**, and the partial install is cleaned.

A new product added to `knownProducts` fails the test until it gets a row, so a fourth feed cannot
land without baseline coverage.

Verified as a real guard by mutation: reverting `ideBuildMatches` to the pre-#429 equality fails it,
and so does flipping Android Studio's `buildIsBaseline` back to `false`.

## Opt-in live tests

Vendor-feed calls stay out of every default run so an outage cannot redden a build:

- `:npx-kt:liveNetworkTest` (`live-network`) — all nine products resolve off the real feeds with a
  plugin-compatible build and a truthful `buildIsBaseline`. ~1 min, no archive download.
- `:npx-kt:liveDownloadSmokeTest` (`live-download`) — `idea-community` and `android-studio` really
  download, unpack and pass validation. Multi-GB, the only thing that reproduces #423 end to end.

Both were run green on macOS arm64 against the live feeds.

## Production change

The feed dispatch was duplicated in `DefaultManagedBackendDownloader.resolve` and
`ReleaseServiceAvailableBackendVersionResolver`; it is now one `resolveBackendArchive` /
`backendDownloadFeed` pair in `BackendDownloadFeeds.kt`, so the download path and the
`devrig backend download` list cannot drift. Behaviour is unchanged.
`resolveArchiveFromProductsApiPayload` becomes public — the same pure-resolver seam its two sibling
feeds already expose.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -229,6 +229,17 @@ Root `build.gradle.kts` defines `ci`-prefixed aggregator tasks for TeamCity and 
 `:test-integration:test` and `:test-experiments:test` have an `onlyIf` guard — plain root `./gradlew test`
 silently skips both. Direct `./gradlew :test-integration:test --tests '...'` still works.
 
+**Vendor-feed tests are opt-in**, so a Google/JetBrains/GitHub outage can never redden a normal build:
+
+| Task | Covers | Cost |
+|------|--------|------|
+| `:npx-kt:liveNetworkTest` | every supported IDE resolves off its live feed with a plugin-compatible build (`live-network` tag) | ~1 min, no archive download |
+| `:npx-kt:liveDownloadSmokeTest` | `idea-community` + `android-studio` really download, unpack and pass `product-info.json` validation (`live-download` tag) | multi-GB per case |
+| `:intellij-downloader:liveNetworkTest` | products-API filename tokens still match (JUnit4 `LiveNetwork` category) | seconds |
+
+The offline equivalent runs by default: `AllIdeProductsDownloadTest` walks all nine products through
+recorded payloads of all three feeds.
+
 **TeamCity DSL** lives in a separate repo (`~/Work/mcp-steroid-teamcity`). See its own `CLAUDE.md` for the
 generate→edit→regenerate→commit workflow. The TC VCS root pulls from `jb`, not `origin` — see "Git remotes" below.
 
--- a/intellij-downloader/src/buildsrc-shared/kotlin/com/jonnyzzz/mcpSteroid/ideDownloader/IdeReleaseLookup.kt
+++ b/intellij-downloader/src/buildsrc-shared/kotlin/com/jonnyzzz/mcpSteroid/ideDownloader/IdeReleaseLookup.kt
@@ -118,7 +118,12 @@ private fun resolveArchiveWithUrlReader(
     )
 }
 
-internal fun resolveArchiveFromProductsApiPayload(
+/**
+ * Resolves an archive from a raw `data.services.jetbrains.com/products` payload — pure, no I/O, so
+ * callers can exercise the products-API feed from a fixture. The sibling feeds expose the same seam
+ * (`resolveGithubCommunityArchiveFromReleasesJson`, `resolveAndroidStudioCanaryArchiveFromHtml`).
+ */
+fun resolveArchiveFromProductsApiPayload(
     product: IdeProduct,
     channel: IdeChannel,
     os: HostOs,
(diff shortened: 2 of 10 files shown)


jonnyzzz:
## Review 2026-08-03

- Staleness: branch sits exactly on main HEAD (30ef5a31), MERGEABLE/CLEAN, CI green. #429 is merged and this extends its coverage honestly: the all-nine-products offline matrix + `knownProducts`-set guard is a real regression fence (mutation-verified per PR body, and the assertions do route through production `ideBuildMatches`/`validateInstalledBuildNumber`). The `BackendDownloadFeeds.kt` dedup is behavior-preserving vs the two removed dispatch copies in `ManagedBackend.kt`/`BackendDownloadListCommand.kt` (same products, same channels, same default arch).
- Cross-PR: no semantic conflict with open #432 — it keeps the `resolveAndroidStudioCanaryArchiveFromHtml(html, os, architecture, version)` seam intact, and this PR's single-generation `androidStudioPreviewHtml` fixture (version-in-URL `install/<v>/...`) still parses under #432's URL-based `resolveAndroidStudioPageDownload`. Whichever merges second just recompiles.
- Convention violation: `npx-kt/src/test/kotlin/com/jonnyzzz/mcpSteroid/devrig/BackendDownloadSmokeTest.kt:57` — `assumeTrue(resolveHostOs() != HostOs.WINDOWS, ...)` is the banned runtime detect-and-skip (CLAUDE.md: only acceptable skip is Gradle task level `enabled = !condition`; the sole `assumeTrue` exception on main is the documented Gemini-key case). Both methods in the class skip on Windows, so gate the `liveDownloadSmokeTest` task itself in `npx-kt/build.gradle.kts` (or wire the installDist 7z.exe so Windows actually runs).
- Nit: `BackendDownloadSmokeTest.kt:81` re-resolves the live feed after the multi-GB download; a release published mid-test fails the final `ideBuildMatches` assert spuriously. Capturing the resolution once (before `manager.download`) closes the TOCTOU window.
- Coverage-honesty note (fine as-is, just naming it): in the default-run `AllIdeProductsDownloadTest`, `downloadAndUnpack` is a fixture (`FeedResolutionDownloader`) — the real transfer/unpack lives only behind `live-download`, which the KDoc discloses.

**Verdict: needs-changes** — content and base are otherwise merge-ready, but the `assumeTrue` runtime skip must move to the Gradle task level per the repo's banned-patterns rule (one-line fix).

jonnyzzz:
## Processed 2026-08-03

**Changes made on the branch before merge:**
- Rebased cleanly onto `origin/main` (no conflicts).
- Fixed the one blocker: `BackendDownloadSmokeTest` used `Assumptions.assumeTrue(resolveHostOs() != HostOs.WINDOWS)` — the repo-banned runtime detection-and-skip. The Windows exclusion now lives at the Gradle task level: `:npx-kt:liveDownloadSmokeTest` is gated `enabled = !isWindows` (+ matching `onlyIf`) in `npx-kt/build.gradle.kts`, following the `ocr-tesseract` precedent. The test methods themselves are now unconditional; the KDoc documents where the gate lives.

**Tests run (unit tasks only, sequentially):**
- `./gradlew :npx-kt:test` — BUILD SUCCESSFUL (includes the new offline `AllIdeProductsDownloadTest` over recorded feed payloads; `live-network`/`live-download` tags excluded by design)
- `./gradlew :intellij-downloader:test` — BUILD SUCCESSFUL
- `./gradlew :npx-kt:liveDownloadSmokeTest --dry-run` — task configuration validated without downloading

The live download smoke itself (multi-GB, `liveDownloadSmokeTest`) was intentionally NOT run here — that path runs on TC.

**Merged** via rebase into `main` (bd4ae141), branch deleted. All PR checks (compile, linux-docker, windows-native) passed on the final head.

PR #432, opened 2026-08-03
devrig: resolve Android Studio by version, not by HTML order

jonnyzzz:
## Problem

`developer.android.com/studio/preview` serves **several generations at once**. Right now:

| Version | Channel | Filename |
|---|---|---|
| `2026.1.4.3` | Canary 3 | `android-studio-quail4-canary3-*` |
| `2026.1.3.6` | RC 2 | `android-studio-quail3-rc2-*` |

`resolveAndroidStudioCanaryArchiveFromHtml` pulled every `android-studio-*.{dmg,exe,tar.gz,zip}` URL
out of the page and took **the first one matching the host suffix**:

```kotlin
val url = urls.firstOrNull { it.endsWith(suffix) }
```

So the resolved build is whichever generation Google happened to type first. Today the canary block
comes first and the result is correct — the day the sections are swapped devrig silently installs the
older RC, with no error and no signal. `AndroidStudioReleaseLookup` (the stable `/studio` page) has
the same first-match-wins shape.

Found while checking `#431`, which added Android Studio to the download coverage added by `#429`:
that test pins the build shape, not which of the page's versions is chosen.

Version pinning was rejected outright on both paths ("version pinning is not supported"), even though
every other feed supports it.

## Fix

New `AndroidStudioDownloadPage.kt` (in `:intellij-downloader`, shared by both pages) parses the page
into **all** the releases it advertises — version from the download URL path, channel from the
filename (`-canary3-`, `-rc2-`, `-beta1-`, none = release), every artifact link — and selects by
comparing version numbers:

- no request → the newest release that publishes an artifact for this OS/arch (falls through to the
  next-newest if a generation lacks it);
- `2026.1.4.3` → that exact version;
- `2026.1` → newest patch of that generation (segment boundary, so `2026.1` never matches `2026.10`);
- `261` → newest build on that platform baseline (the artifact then reports `261.26222.65.…`);
- `canary` / `rc` / `beta` → newest build of that channel.

Channel never overrides the version: a same-version tie breaks toward the more stable channel, so the
order is a total function of the page content. This keeps today's behaviour (canary 2026.1.4.3 wins)
while making it independent of Google's layout.

`resolveAndroidStudioArchiveFromHtml` (stable page) now routes through the same selector and keeps its
SHA-256 table lookup; `inferAndroidStudioVersion` became a thin wrapper over the shared regexes.

## Tests

`AndroidStudioCanaryReleasesTest` gains the **recorded** preview page (2026-08-03, trimmed to the
download-button blocks) plus the same page with the two generations **swapped** so the older RC comes
first — the layout that used to resolve wrong. It asserts the canary still wins on mac/linux/windows,
and covers exact / prefix / baseline / channel pinning and the error message that lists what the page
does offer. `AndroidStudioDownloadPageTest` covers the parser and selector directly.

Verified as a real guard by mutation: dropping the newest-first sort fails
`the html order of the release blocks does not decide which build is picked` and
`a platform baseline request resolves to the newest build on that baseline` (and 2 in the downloader
suite); restoring it makes them pass.

Green: `./gradlew :npx-kt:test :intellij-downloader:test` — 1404 + 103 tests, 0 failures.

Real download on macOS arm64: `devrig backend download android-studio` resolved `2026.1.4.3` /
baseline `261`, fetched `android-studio-quail4-canary3-mac_arm.dmg`, unpacked and validated — the
install reports `261.26222.65.2614.15978069`, matching the baseline. Artifacts deleted afterwards.

## Not in this PR

`IdeChannel.EAP` for Android Studio in `:intellij-downloader` is still unimplemented (only `STABLE` is
accepted), so `IdeDistribution.Latest(AndroidStudio, EAP, "2026.1")` in `AndroidStudio261Test` still
cannot resolve. The shared parser is what that would need; wiring it is a separate change.

### DIFF
--- a/intellij-downloader/src/buildsrc-shared/kotlin/com/jonnyzzz/mcpSteroid/ideDownloader/AndroidStudioDownloadPage.kt
+++ b/intellij-downloader/src/buildsrc-shared/kotlin/com/jonnyzzz/mcpSteroid/ideDownloader/AndroidStudioDownloadPage.kt
@@ -1,0 +1,174 @@
+/* Copyright 2025-2026 Eugene Petrenko (mcp@jonnyzzz.com); Copyright 2025-2026 JetBrains. Use of this source code is governed by the Apache 2.0 license. */
+package com.jonnyzzz.mcpSteroid.ideDownloader
+
+// Parses a `developer.android.com` download page — `/studio` (stable) or `/studio/preview` (canary,
+// beta, RC) — into ALL the releases it advertises, and picks one of them deterministically.
+//
+// Google serves several generations from the same page: the preview page currently offers
+// `2026.1.4.3` (Quail 4 Canary 3) next to the older `2026.1.3.6` (Quail 3 RC 2). Which block comes
+// first in the HTML is Google's editorial choice, so taking the first matching download link resolves
+// whatever happens to be typed first — correct today, silently the older RC the day Google swaps the
+// sections. Selection here reads only the page CONTENT: links are grouped by the version in their
+// URL, and the pick is made by comparing version numbers.
+
+/** Release channel of an Android Studio build, as spelled in its download filename. */
+enum class AndroidStudioPageChannel(val token: String, val stability: Int) {
+    /** No channel segment (`android-studio-quail3-linux.tar.gz`) or a stable patch (`-panda4-patch1-`). */
+    RELEASE("release", 3),
+    RC("rc", 2),
+    BETA("beta", 1),
+    CANARY("canary", 0),
+}
+
+/** One Android Studio version as published on a download page, with every artifact link it lists. */
+data class AndroidStudioPageRelease(
+    /** Marketing version taken from the download URL, e.g. `2026.1.4.3`. */
+    val version: String,
+    val channel: AndroidStudioPageChannel,
+    /** Download URLs published for this version, in page order. */
+    val urls: List<String>,
+) {
+    /** The IntelliJ platform baseline this build is made of: `2026.1.4.3` -> 261. */
+    val platformBaseline: Int? get() = androidStudioPlatformBaseline(version)
+
+    fun assetEndingWith(suffix: String): String? = urls.firstOrNull { it.endsWith(suffix) }
+}
+
+/** The artifact chosen for one host out of an [AndroidStudioPageRelease]. */
+data class AndroidStudioPageDownload(
+    val release: AndroidStudioPageRelease,
+    val url: String,
+)
+
+/**
+ * Maps an Android Studio marketing version to its IntelliJ platform baseline: `2026.1.2.3` -> 261,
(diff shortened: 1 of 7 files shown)


jonnyzzz:
## Review 2026-08-03

- Selector core verified (`AndroidStudioDownloadPage.kt`): `matchesSelector` handles exact version / segment-boundary prefix (`2026.1` cannot match `2026.10.x`) / 100..999 baseline / channel; `compareAndroidStudioVersions` is segment-wise; the comparator tie-break chain (version → channel stability → raw text) makes selection a total function of page content as claimed. No counterexample found.
- Test coverage is honest: nothing removed or weakened; the swapped fixture self-guards (`indexOf` assertion at AndroidStudioCanaryReleasesTest.kt:79) so it cannot silently stop guarding. The empty-page throw-type change (`require`/IAE → `error`/ISE) breaks no prior assertion — the only pre-existing throw-type test (linux-arm64) still routes through the unchanged `require` in `androidStudioPreviewSuffix`.
- Cross-module move of `androidStudioPlatformBaseline` (npx-kt → intellij-downloader) is safe: the only consumer outside touched files is `IdeReleaseLookupTest.kt:327` via the retained `inferAndroidStudioVersion` wrapper, and the compile-check workflow compiles every source set (passed).
- Staleness: none. Merge-base dccc19b1 is 1 commit behind HEAD 30ef5a31 (test-integration only); mergeable CLEAN. Open #431 (which surfaced this bug) has zero file overlap — complementary, no ordering constraint.
- Nit (non-blocking), AndroidStudioDownloadPageTest.kt:129-134: the `errorMessageOf` helper with `catch (e: Throwable)` + `@Suppress("UNREACHABLE_CODE")` scaffolding could just be `assertThrows<IllegalStateException>{...}.message` as the npx-kt test already does.

**Verdict: ready** — deterministic version-based selection replaces HTML-order luck, version/baseline/channel pinning now works on both pages (previously rejected outright), fixtures are recorded and adversarially swapped, and nothing on recent main conflicts.

jonnyzzz:
## Processed 2026-08-03

**Branch changes:** rebased onto current main (`30ef5a31`) — clean rebase, no conflicts. No code fix was needed (the 2026-08-03 review found the PR defect-free); the single PR commit is unchanged.

**Tests run** (sequentially, in an isolated worktree):
- `./gradlew :intellij-downloader:test` — BUILD SUCCESSFUL
- `./gradlew :npx-kt:test` — BUILD SUCCESSFUL

**Final review:** diff vs main confirmed minimal and on-topic (shared `AndroidStudioDownloadPage` parser/selector, both resolvers routed through it, tests + recorded preview-page fixtures); no banned patterns.

**Merged** via rebase as `91f89d9a`; branch deleted. All GitHub checks (compile, installer linux-docker, installer windows-native) passed on the rebased head.

jonnyzzz:

Issue #412, opened 2026-07-31
test-integration: IDE smoke matrix newly unblocked by #408 — 11 latent failures surfaced (classpath drift, AS readiness, managed-backend, container cluster)


jonnyzzz:
Fixing #408 (Gemini missing-key now SKIPPED instead of 6 hard failures) unblocked the `ciIntegrationTests` chain: `:ij-plugin:integrationTest` no longer fails the build, so `:test-integration:test` — the Docker IDE smoke matrix — **executes on TC for the first time in months** (TestIntegrationBuild run 1019004081 on jb-merge `7ecf8686`: 225+ passed, 10 ignored as designed, 11 failed).

The 11 failures are latent drift accumulated while the matrix was dark, in four buckets:

1. **`ContentModuleClasspathTest.ideClasspath covers all JARs`** — 41 JAR(s) on filesystem not in classpath and not covered by any exception rule (`lib/intellij.rd.client*.jar`, DatabaseTools/Kotlin completion-ML modules, docker compose frontend, …). The rolling IDE build moved ahead of `UNLOADED_CONTENT_MODULES` — same rolling-channel-slide mechanism as the audit's GAP-2 (`TODO-TC-COVERAGE-AUDIT.md`).
2. **`AndroidStudioRuntimeCompatTest.runtime compat android studio 261`** — "Failed waiting for MCP Steroid server ready after 301365ms": AS container readiness exceeded the 5-min budget on the TC agent (cold image / load). May be infra; needs a rerun to classify.
3. **`DevrigManagedBackendGuiIntegrationTest`** — "Process download IDEA Community managed backend failed" on devrig `0.101.599-jb-7ecf868`; output shows the jb-lane launcher self-heal narration. Needs log dig: real download failure vs lane-dependent output parsing (cf. the CliOptionsIntegrationTest lane fix, commit 499942e6).
4. **Container/env cluster** — `IntelliJContainerTest` ×4 (incl. "container starts and IDE becomes ready", xdotool), `GradleCompileTest`, `GradleTestExecutionTest`, `IdeaPlaygroundTest`, `JdkTableIntegrationTest`. Broad IDE-container failures, likely shared root cause (image/agent state); triage from the run artifacts (screenshots + thread dumps per `test-integration/AGENTS.md`).

None of these are regressions from the #406–#410 fixes (unrelated classes; the matrix simply never ran on the previous merges — the earlier "108 passed" runs died at the Gemini step before reaching it).

[...]

PR #433, opened 2026-08-03
Move execution history to shared collision-safe storage

### LINKED ISSUE #280: Refactor ExecutionStorage to meep all files off the project
Our current implementation is to keep files under the `.idea/mcp-steroid`, 
instead we should place all the run folders under the devrig home,
which is `~/.mcp-steroid/runs`

relates to #73 #261 #78

### LINKED ISSUE #414: Execution IDs collide: two concurrent executions share one storage directory
`ExecutionStorage.newExecutionId` can return the same ID for two executions, and nothing detects it —
they silently merge into one directory.

## Reproduction

Concurrently, in the same second:

```bash
devrig execute_code    --task_id=screenshot-cross-b --code='println("x")' --reason=r
devrig take_screenshot --task_id=cross-b --reason=r
```

Both get `eid_20260802T205049-screenshot-cross-b`: the screenshot handler prefixes its own key
(`taskId = "screenshot-$taskId"`, `VisionScreenshotToolHandler.kt:31`), so both storage keys match. The
directory ends up holding the screenshot *and* the compiled script, while `tool.json` describes only the
screenshot.

## Root cause

`execution-storage/.../ExecutionStorage.kt:175` — ID is `eid_` + `yyyyMMdd'T'HHmmss` + normalised
`task_id`, with no uniqueness check:

- second precision;
- any run outside `[a-zA-Z0-9_-]` collapses to one `_`, so `α`/`β` and `a.b`/`a,b`/`a b` all collide;
- `LocalDateTime.now()` is local, so a DST fall-back repeats an hour of IDs.

## Impact

`tool.json`, `script.kts`, `reason.txt` overwritten by the last writer; `output.jsonl` interleaved;
`execution_id` no longer identifies one execution. Execution itself is fine — compilation is serialised
and each client gets its own response. Low as security/runtime, high as audit and diagnostics. Worth
fixing because the CLI encourages reusing one `task_id` across related calls.

Also the cause of "four concurrent `execute_feedback` calls, four `ACK!`s, two persisted".

## Fix

Make `task_id` a decorative slug; take uniqueness from a UUID or atomic directory reservation. If a
timestamp stays, use UTC with milliseconds. Keep the original `task_id` identical across
code/screenshot/input/feedback. Don't let multiple screenshots overwrite each other.

Nearby: `findExecutionId` (`:156`) rejects `/` and `..` but not `\` or absolute paths — traversal gap on
Windows.

**Acceptance:** eight concurrent calls sharing one `task_id`, including Unicode and a cross-tool pair,
produce eight distinct directories.

Found on devrig `0.101.19999-SNAPSHOT` / IntelliJ IDEA 2026.1.3 while exercising #284; storage code
untouched by that work.

jonnyzzz:
## Summary

- move execution history from project `.idea` folders to `~/.mcp-steroid/runs` and update active documentation/configuration
- reserve execution directories atomically with UTC time, an epoch-millisecond modulo-997 slot, bounded project/task slugs, and one-character backend provenance
- persist the full backend identity in `backend_name.txt`, preserve original task IDs across tool types, and harden execution-ID path validation
- stamp devrig provenance authoritatively inside the IDE bridge while keeping persisted tool JSON and older devrig clients compatible

## Verification

- `:mcp-core:test` — 121 tests, 0 failures
- `:execution-storage:test` — 16 tests, 0 failures
- `:mcp-steroid-server:test` — 83 tests, 0 failures
- `:npx-kt:test` — 1,397 tests, 4 existing skips, 0 failures
- `:ij-plugin:test` — 280 tests, 1 existing skip, 0 failures
- `DevrigManagedBackendGuiIntegrationTest` — 1 test, 0 failures
- IntelliJ inspections ran on every changed production Kotlin file; new/modified code is clean (baseline-only findings remain in untouched declarations)
- `git diff --check`

Closes #280
Closes #414

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -55,7 +55,7 @@ When reporting a bug, please include:
 - Steps to reproduce the issue
 - Expected vs. actual behavior
 - IDE version, OS, and MCP Steroid version
-- Relevant log snippets (from `.idea/mcp-steroid/` if applicable)
+- Relevant log snippets (from `~/.mcp-steroid/runs/` if applicable)
 
 Feature requests should describe the use case and motivation, not just the desired solution.
 
--- a/README.md
+++ b/README.md
@@ -234,7 +234,7 @@ MCP Steroid can be configured via IntelliJ's Registry (`Help > Find Action > Reg
 |--------------|---------|-------------|
 | `mcp.steroid.server.port` | 6315 | MCP server port (0 for auto-assign) |
 | `mcp.steroid.server.host` | 127.0.0.1 | Bind address (use 0.0.0.0 for Docker) |
-| `mcp.steroid.storage.path` | (empty) | Custom storage path (default: .idea/mcp-steroid/) |
+| `mcp.steroid.storage.path` | (empty) | Custom storage path (default: `~/.mcp-steroid/runs/`) |
 
 See the full [Configuration Documentation](https://devrig.dev/docs/configuration/) on the website.
 
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -11,7 +11,7 @@ This document is a concise architecture map. For authoritative details, see `AGE
 - Vision tools: screenshot/input tooling with artifact storage.
 - OCR helper: external `ocr-tesseract` app invoked via process client.
 - Kotlinc helper: bundled Kotlin compiler invoked via process client.
-- Storage: execution logs/artifacts (append-only, under `.idea/mcp-steroid/`).
+- Storage: execution logs/artifacts (append-only, under `~/.mcp-steroid/runs/` by default).
 - **devrig CLI** (`npx-kt/`): stateless stdio MCP server + `backend` /
   `project` CLI that discovers IntelliJ instances on the host and
   routes tool calls to them. Project / IDE naming is governed by the
--- a/execution-storage/src/main/kotlin/com/jonnyzzz/mcpSteroid/storage/ExecutionStorage.kt
+++ b/execution-storage/src/main/kotlin/com/jonnyzzz/mcpSteroid/storage/ExecutionStorage.kt
@@ -2,6 +2,7 @@
 package com.jonnyzzz.mcpSteroid.storage
 
 import com.jonnyzzz.mcpSteroid.server.ExecCodeParams
+import com.jonnyzzz.mcpSteroid.server.ExecutionBackendProvenance
 import com.jonnyzzz.mcpSteroid.server.FeedbackParams
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
@@ -11,12 +12,13 @@ import kotlinx.serialization.json.JsonObject
 import kotlinx.serialization.json.buildJsonObject
 import kotlinx.serialization.json.encodeToJsonElement
 import kotlinx.serialization.json.jsonObject
+import java.nio.file.FileAlreadyExistsException
 import java.nio.file.Files
 import java.nio.file.Path
 import java.nio.file.StandardOpenOption
-import java.time.LocalDateTime
+import java.time.Clock
+import java.time.ZoneOffset
 import java.time.format.DateTimeFormatter
-import kotlin.io.path.deleteIfExists
 import kotlin.io.path.isRegularFile
 import kotlin.io.path.writeText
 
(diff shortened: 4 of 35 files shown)


jonnyzzz:
## Review 2026-08-03

- **Core fix verified sound.** Atomic `Files.createDirectory` reservation + UTC clock + full-match `SAFE_EXECUTION_ID` (execution-storage/ExecutionStorage.kt) resolves #414's collision, DST, and Windows-`\` traversal gaps; the 8-way concurrent unicode/cross-tool test under a fixed clock matches the issue's acceptance criterion honestly. Provenance is unspoofable: `executionBackendProvenance()` reads only `@Transient trustedArguments`, which JSON deserialization can never populate — covered on both the bridge (NpxBridgeServiceTest) and server (ExecutionBackendProvenanceTest) sides. #280 is delivered (StoragePaths default, HomePaths `runs`, vmoptions un-pinned with an absence assertion in DevrigManagedBackendGuiIntegrationTest).
- **Defect — unit tests now pollute the real `~/.mcp-steroid/runs`.** Only ExecutionStorageTaskTest.kt:33-34 gets the `mcp.steroid.storage.path` isolation override; the ~dozen other execution-running ij-plugin test classes (ScriptExecutorTest, ExecutionManagerTest, DialogKillerTest, LspExamplesExecutionTest, McpServerIntegrationTest, …) hit the new StoragePaths.kt:53 default and write append-only execution folders (script.kts, compiled artifacts, output.jsonl) into the developer's / TC agent's real home — never cleaned, growing every build, and contaminating the very `~/.mcp-steroid/runs` corpus this PR's docs ask users to share for telemetry. Hoist the registry override into the shared setup hook (BaseTestCase.kt:19 `setServerPortProperties()` is already called from these tests' `setUp`).
- Minor: `newExecutionId`'s retry loop is unbounded if the injected `Clock` never advances once all 997 slots for a second are taken (unreachable with `systemUTC`, but a bounded-attempts `check` would make the loop total, and the bare `catch (FileAlreadyExistsException) { attempt++ }` deserves a one-line comment that the exception IS the collision detector).
- Minor: `tool.json` `timestamp` silently changes format from local `ISO_LOCAL_DATE_TIME` to `Instant.toString()` (UTC `Z`) — an improvement and no in-repo consumer parses it, but it contradicts the PR body's "persisted tool JSON compatible" claim; worth a line in the description.
- Nit: `ExecutionBackendInfo` (execution-storage) duplicates `ExecutionBackendProvenance` (mcp-steroid-server) — identical fields and `init` invariants, and execution-storage already imports the server class; one type would do.
- Staleness: base is main~1 (merge-base dccc19b1), MERGEABLE/CLEAN; the only main movement since is the #412 hotfix 30ef5a31 touching test-integration infra files this PR doesn't — no textual or semantic conflict. Debugger prompt recipes writing `.idea/mcp-steroid/debug-events.ndjson` self-`mkdirs()` and are unaffected.

**Verdict: needs-changes** — the storage move and collision fix are correct and well-tested, but the test-suite must not write into the real user home; one small shared-setup change (plus the optional minor notes) and this is ready.

jonnyzzz:
## Processed 2026-08-03

**Changes made on the branch before merge:**
- Rebased onto `origin/main` (clean, no conflicts).
- Fixed the review blocker (test-storage isolation was per-class only): `ij-plugin/build.gradle.kts` now passes `-Dmcp.steroid.storage.path=<build/mcp-steroid-test-runs/<task>-<timestamp>>` to every `Test` task JVM (`test`, `integrationTest`, future ones) via `tasks.withType<Test>` + `CommandLineArgumentProvider`. The registry resolves stored user values first, then `System.getProperty(key)`, then the bundled default (verified against `RegistryValue.resolveNotRequiredValue` in the 2026.1 platform), so one JVM-level property isolates ALL execution-running test classes without per-class setUp code — and, being an argument provider, it is not fingerprinted as a task input, so build caching stays intact.
- Removed the now-redundant per-class registry override from `ExecutionStorageTaskTest`; rewrote `testDefaultStorageLivesUnderMcpSteroidHome` to clear the isolation property for its duration (the platform drops a stored value equal to the bundled default, which would otherwise fall through to the property); added `testExecutionStorageIsIsolatedFromRealUserHome` to pin the contract.

**Tests run (worktree, sequential):**
- `:execution-storage:test` — pass
- `:mcp-core:test` — pass
- `:mcp-steroid-server:test` — pass
- `:npx-kt:test` — pass
- `:ij-plugin:test` — 281 tests, 0 failures, 1 pre-existing skip — pass

**Acceptance proof for the fix:** `ls ~/.mcp-steroid/runs` before vs after the full `:ij-plugin:test` run — identical (431 entries, no stray folders). The 46+ execution folders produced by the suite (25 MB) all landed in `ij-plugin/build/mcp-steroid-test-runs/test-<timestamp>/` instead.

Merged into `main` via rebase (`0748c941`); branch deleted. Non-blockers from the review (tool.json timestamp format, unbounded retry under a fixed Clock) were intentionally left out of scope.

PR #434, opened 2026-08-03
devrig: add PhpStorm, RubyMine, DataGrip, RustRover and MPS to the IDE product catalog

### LINKED ISSUE #430: devrig: show and support the full JetBrains IDE product set, not a hard-coded nine
## Problem

devrig only knows nine IDE products. Everything user-facing — `devrig backend download`
(the list and the download), the docs table, the `--product` flag of the standalone
downloader — is derived from a single hard-coded list, so the whole product presents
itself as "these nine IDEs" rather than "JetBrains IDEs".

That contradicts what we promise elsewhere. `README.md:74` says "IntelliJ IDEA 2026.1+
(or any IntelliJ-based IDE: Rider, Android Studio, GoLand, WebStorm, PyCharm, CLion,
etc.)" — the "etc." has no implementation behind it. A PhpStorm or RubyMine user who runs
`devrig backend download` sees no PhpStorm and no RubyMine, and asking for one by name is
a hard error, not a fallback.

## Where the list lives today

Single source of truth:

- `intellij-downloader/src/buildsrc-shared/kotlin/com/jonnyzzz/mcpSteroid/ideDownloader/IdeProduct.kt:182`
  — `IdeProduct.Companion.knownProducts`, nine entries.
- Same file, `knownByAlias` (line 196) + `fromString` (line 236): unknown ids fail with
  "Unsupported IDE product '…'".

Everything user-facing reads from it:

- `npx-kt/.../devrig/BackendDownloadListCommand.kt:127` — `orderedKnownBackendProducts()`,
  i.e. the rows printed by `devrig backend download` with no id.
- `npx-kt/.../devrig/BackendId.kt:101` — `parseManagedProductKeyOrNull` matches
  `knownProducts` only, so an unknown id is rejected before any feed is queried.
- `npx-kt/.../devrig/BackendLocalListCommand.kt:152`, `.../devrig/BackendCommand.kt:284`
  — local/installed listing and ordering.
- `website/content/docs/devrig.md:99-100` — the published "Known product ids" line.
- `intellij-downloader/.../ideDownloader/Main.kt:16-18` — the `--product` help text.
- `test-integration/.../integration/infra/ide-product.kt:92` — the integration matrix,
  seven products.

## Current vs missing

Supported today (9): `idea-ultimate` (IIU), `idea-community` (IIC), `pycharm-pro` (PCP),
`pycharm-community` (PCC), `goland` (GO), `webstorm` (WS), `rider` (RD), `clion` (CL),
`android-studio` (AI).

Missing, all served by the same `data.services.jetbrains.com/products?code=…` feed the
resolver already uses (codes verified against the live API):

| Product | Code |
|---|---|
| PhpStorm | `PS` |
| RubyMine | `RM` |
| DataGrip | `DG` |
| RustRover | `RR` |
| DataSpell | `DS` |
| MPS | `MPS` |
| Gateway | `GW` |

Adding one is mechanical: an `id` / `displayName` / `code` / `launcherExecutable` /
`licenseTier` / `urlFilenameTokens` block plus a `knownProducts` entry. No new feed and no
new resolution strategy — unlike Android Studio, which needed its own scraper.

## The inconsistency is already inside the repo

The prompt pipeline is ahead of the downloader. `prompts/AGENTS.md:157-159` documents
`RM` (RubyMine) and `DB` (DataGrip) as first-class product codes, and
`prompts/src/test/kotlin/.../PerIdeAvailabilityContractTest.kt:38` asserts availability
across `IU, PY, GO, WS, RD, CL, RM, DB`. So we ship prompts for RubyMine and DataGrip that
devrig cannot download a backend for. (Note the code spelling differs too — the prompt
pipeline uses `DB` for DataGrip, the products API uses `DG`; whatever we add should
reconcile that rather than introduce a third spelling.)

`IdeProduct.Custom` exists and its KDoc even names `RR` / `AC` / `DG` as examples, but
nothing user-facing can reach it: `BackendId.parseManagedProductKeyOrNull` and
`IdeProduct.fromString` both consult `knownProducts` only. `Custom` is API-only today.

## Proposal

- Extend `IdeProduct.knownProducts` to the full IDE product set, with per-product
  `urlFilenameTokens` and `installedProductCode` verified against real feed payloads
  (the existing fixture pattern, e.g. `intellij-downloader/src/test/resources/fixtures/products-WS.json`).
- Decide and document the boundary: which JetBrains products are in scope as *devrig
  backends* (IntelliJ-platform IDEs) and which are explicitly out (ReSharper, Qodana,
  dotTrace, Toolbox, …). Gateway and MPS are edge cases worth an explicit yes/no.
- Make the product catalog the single source for every surface — CLI list, `--product`
  help, `website/content/docs/devrig.md`, the integration matrix — so a new product does
  not need six edits and cannot drift.
- Reconcile the prompt pipeline codes (`RM`, `DB`) with the downloader catalog so a
  product we ship prompts for is a product devrig can provision.
- Either expose a documented escape hatch for a product code we do not ship an alias for,
  or state plainly in the error message that the catalog is the whole supported set.

## Acceptance criteria

- [ ] `devrig backend download` with no id lists the full agreed IDE set, grouped by
      license tier as today.
- [ ] Each newly added product resolves a real archive for macOS/Linux/Windows × x64/arm64
      (per-product filename tokens verified, not guessed).
- [ ] `website/content/docs/devrig.md` and the `--product` help are generated from or
      checked against `knownProducts` — a test fails if they drift.
- [ ] The in/out-of-scope decision is written down in the repo, not only in this issue.
- [ ] Every product code the prompt pipeline claims to support has a downloadable backend,
      or is explicitly documented as prompts-only.

## Context

Follow-up to the release-resolution work: #429 (baseline-only build acceptance for
`idea-community` downloads) fixed how one product's version is compared, and the
#416–#427 clean-machine set is about an agent finding and using `devrig backend download`
at all. Both assume the product list in front of the user is the real one. Today it is a
subset.

jonnyzzz:
Fixes #430

devrig only knew nine IDEs, so `devrig backend download` presented itself as "these nine" and asking
for a PhpStorm or RubyMine backend by name was a hard error. This extends the product catalog to
fourteen.

## Added (5)

| Product | id | API code | `product-info.json` code | License tier |
|---|---|---|---|---|
| PhpStorm | `phpstorm` | `PS` | `PS` | paid |
| RubyMine | `rubymine` | `RM` | `RM` | paid |
| DataGrip | `datagrip` | `DG` | **`DB`** | paid |
| RustRover | `rustrover` | `RR` | `RR` | free for non-commercial |
| MPS | `mps` | `MPS` | `MPS` | free |

**DataGrip's two codes are deliberate.** The products feed is queried with `DG`
(`alternativeCodes: ["DB"]`), but its `intellijProductCode` is `DB`, so a real install writes
`productCode: DB` into `product-info.json`. `installedProductCode` carries `DB` — the same split
already handled for `IIU`→`IU` and `PCP`→`PY` — and the alias map accepts both spellings. `DB` is
also the code `prompts/AGENTS.md` and `PerIdeAvailabilityContractTest` use, so this closes the
prompts-ship-for-it/devrig-cannot-download-it mismatch the issue described (a test now asserts every
prompt-pipeline code has a downloadable backend).

**MPS is the odd feed entry.** It publishes no `linuxARM64` and no `windowsARM64` distribution in any
release. It does offer a cross-platform `zip`, which we deliberately do **not** fall back to: that
archive carries no per-host JBR, so "resolved" would mean an install that cannot launch. Instead the
resolver now separates "the feed publishes no such platform" from "filenames belong to another
edition" and reports the download keys the feed does offer, rather than advising a `--version` retry
that cannot help. Its `mac`/`macM1` `.dmg` needs no special handling — `unpackIdeArchive` already
mounts DMGs via `hdiutil`.

## Deliberately NOT added

Recorded as code comments next to `IdeProduct.knownProducts` so this does not read as an oversight to
a future reader, and guarded by a test that fails if one is added without revisiting the decision:

- **DataSpell (`DS`)** — consciously left out by decision.
- **GitClient (`GIG`)** — consciously left out by decision.
- **CLion Nova (`CLN`)** — dead upstream: the feed reports `alternativeCodes: [CL]` and
  `intellijProductCode: CL`, i.e. folded back into regular CLion, which we already ship as `CL`. A
  separate entry would duplicate the same artifacts.
- **Gateway (`GW`)** — a thin remote-development client, not a backend IDE.
- **JetBrains Air / Fleet** — not IntelliJ-platform desktop distributions.
- **AppCode (`AC`)** — discontinued.
- **PhpStorm Light** — a stripped SKU of the `phpstorm` entry now shipped.
- **Aqua** — test-automation IDE, outside the agreed backend scope.
- **Writerside / Qodana** — not IDEs.

`IdeProduct.Custom` remains the escape hatch for any of these without touching the catalog.

## Where it changed

- `IdeProduct.kt` — the five entries, the alias additions (`php`, `ruby`, `rust`, `db`), and the
  out-of-scope decision comment.
- `IdeReleaseLookup.kt` — the unpublished-platform diagnostic.
- `Main.kt` `--product` help, `website/content/docs/devrig.md` "Known product ids".
- `test-integration/.../infra/ide-product.kt` — five matrix entries plus their Dockerfiles;
  `intelliJ-download.kt` product mapping (its exhaustive `when` caught the omission at compile time).
- No edits were needed in `BackendId.kt`, `BackendDownloadListCommand.kt`, `BackendLocalListCommand.kt`
  or `BackendCommand.kt`: all four already derive from `knownProducts`, so the new products flow into
  id parsing, the download list, local listing and ordering for free. The list groups them by license
  tier as before.

## Verification

- **Metadata verified against reality, not guessed.** All five Linux archives were really downloaded
  and their `product-info.json` read: `PS`/`RM`/`RR`/`MPS` report their own code, DataGrip reports
  `DB`, and each `launch.launcherPath` matches the `launcherExecutable` recorded here. MPS's
  `hasJavaSdk = false` was confirmed the same way (it bundles only `mps-kotlin`, no
  `com.intellij.java`).
- **Real end-to-end download** through the production path (resolve → download → unpack → validate)
  for DataGrip and MPS on macOS arm64 (`.dmg` via `hdiutil`): `productCode` and `buildNumber` both
  matched the resolved build exactly. Downloads were cleaned up afterwards.
- **Offline matrix** (`:intellij-downloader:test`) walks every catalog product across all six
  OS/arch combos off recorded fixtures. A new `knownProducts` entry fails until its fixture exists
  and its filename tokens are right, so a product cannot land in the CLI list while being
  undownloadable. MPS's two ARM64 gaps are asserted as explicit unpublished-platform errors.
- **Live feed matrix** (`:intellij-downloader:liveNetworkTest`, opt-in) repeats the same assertions
  against the real products API — green for all 13 feed-served products × 6 combos.
- **Docs drift** — `ProductCatalogDocsDriftTest` fails when the published docs or the `--product`
  help stop listing a catalog id (an acceptance criterion of #430).
- **Mutation-checked, so the guards are real:** flipping DataGrip's filename token to the wrong case
  or its mapping back to `DG` reds three tests; dropping `mps` from `devrig.md` reds the drift test.
- `./gradlew :intellij-downloader:test :npx-kt:test` green (103 + 1397 tests, 0 failures);
  `:test-integration` and `:prompts` compile clean.

`:prompts:test` was not run to completion: its IDE matrix is a hand-maintained list
(`idea`/`rider`/`clion`/`pycharm`), not derived from `knownProducts`, so this diff does not change it,
and the task downloads ~10 GB of IDE distributions.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -237,7 +237,7 @@ silently skips both. Direct `./gradlew :test-integration:test --tests '...'` sti
 | `:npx-kt:liveDownloadSmokeTest` | `idea-community` + `android-studio` really download, unpack and pass `product-info.json` validation (`live-download` tag) | multi-GB per case |
 | `:intellij-downloader:liveNetworkTest` | products-API filename tokens still match (JUnit4 `LiveNetwork` category) | seconds |
 
-The offline equivalent runs by default: `AllIdeProductsDownloadTest` walks all nine products through
+The offline equivalent runs by default: `AllIdeProductsDownloadTest` walks every catalog product through
 recorded payloads of all three feeds.
 
 **TeamCity DSL** lives in a separate repo (`~/Work/mcp-steroid-teamcity`). See its own `CLAUDE.md` for the
--- a/intellij-downloader/build.gradle.kts
+++ b/intellij-downloader/build.gradle.kts
@@ -77,6 +77,9 @@ tasks.test {
     useJUnit {
         excludeCategories("com.jonnyzzz.mcpSteroid.ideDownloader.LiveNetwork")
     }
+    // ProductCatalogDocsDriftTest reads the published docs / CLI help straight from the
+    // working tree, so it needs the repo root rather than the module dir.
+    systemProperty("mcp.repo.root", rootProject.layout.projectDirectory.asFile.absolutePath)
     // OS-specific skip lives inside each test (`Assume.assumeTrue` / `assumeFalse`).
     // SevenZipLocatorTest needs the Windows-bundled resources → Windows-only.
     // IdeUnpackerSecurityTest hits tar path-separator behavior that differs on
--- a/intellij-downloader/src/buildsrc-shared/kotlin/com/jonnyzzz/mcpSteroid/ideDownloader/IdeProduct.kt
+++ b/intellij-downloader/src/buildsrc-shared/kotlin/com/jonnyzzz/mcpSteroid/ideDownloader/IdeProduct.kt
@@ -146,6 +146,84 @@ sealed interface IdeProduct {
         override val urlFilenameTokens = listOf("CLion-")
     }
 
+    /**
+     * RustRover. The products feed serves only the `release` and `eap` channels for `RR` — there is
+     * no `rc` stream — which both [IdeChannel] values already cover.
+     */
+    data object RustRover : IdeProduct {
+        override val id = "rustrover"
+        override val displayName = "RustRover"
+        override val code = "RR"
+        override val launcherExecutable = "rustrover"
+        override val licenseTier = LicenseTier.FreeForNonCommercial
+        // HEAD evidence: RustRover archives use case-sensitive RustRover-* filenames.
+        override val urlFilenameTokens = listOf("RustRover-")
+    }
+
+    // ---- Paid IDEs ----
+
+    data object PhpStorm : IdeProduct {
+        override val id = "phpstorm"
+        override val displayName = "PhpStorm"
+        override val code = "PS"
+        override val launcherExecutable = "phpstorm"
+        override val licenseTier = LicenseTier.Paid
+        // HEAD evidence: PhpStorm archives use case-sensitive PhpStorm-* filenames (served under /webide/).
+        override val urlFilenameTokens = listOf("PhpStorm-")
+    }
+
+    data object RubyMine : IdeProduct {
+        override val id = "rubymine"
+        override val displayName = "RubyMine"
+        override val code = "RM"
+        override val launcherExecutable = "rubymine"
+        override val licenseTier = LicenseTier.Paid
+        // HEAD evidence: RubyMine archives use case-sensitive RubyMine-* filenames (served under /ruby/).
+        override val urlFilenameTokens = listOf("RubyMine-")
+    }
+
+    /**
+     * DataGrip. **Two different codes are in play, deliberately:** the products API is queried with
+     * `DG` (`alternativeCodes: ["DB"]`), while the feed's `intellijProductCode` — and therefore the
+     * `productCode` a real install writes into `product-info.json` — is `DB`. Same split as
+     * `IIU`→`IU` and `PCP`→`PY`, so [installedProductCode] carries `DB` and validation passes.
+     * `DB` is also the spelling `prompts/AGENTS.md` uses, so the alias map accepts both.
+     */
+    data object DataGrip : IdeProduct {
+        override val id = "datagrip"
+        override val displayName = "DataGrip"
+        override val code = "DG"
+        override val launcherExecutable = "datagrip"
+        override val licenseTier = LicenseTier.Paid
+        override val installedProductCode = "DB"
+        // HEAD evidence: DataGrip archives use LOWERCASE datagrip-* filenames, unlike its sibling IDEs.
+        override val urlFilenameTokens = listOf("datagrip-")
+    }
+
+    // ---- Free IDE ----
+
+    /**
+     * MPS (Apache 2.0, free). Its feed entry is the odd one in the catalog:
+     *
+     *  - **no `linuxARM64` and no `windowsARM64` distribution** in any release, so those two hosts
+     *    are simply not published; the resolver reports that explicitly instead of suggesting
+     *    `--version` (see `unpublishedPlatformFailureMessage`).
+     *  - it additionally offers a cross-platform `zip` that the OS/arch → download-key mapping does
+     *    not use. We deliberately do NOT fall back to it: that archive carries no per-host JBR, so
+     *    "resolved" would mean an install that cannot launch — a worse outcome than a clear error.
+     *  - `mac` / `macM1` are `.dmg`, same as every other JetBrains IDE, and `unpackIdeArchive`
+     *    already mounts those via `hdiutil` on a macOS host.
+     */
+    data object Mps : IdeProduct {
+        override val id = "mps"
+        override val displayName = "MPS"
+        override val code = "MPS"
+        override val launcherExecutable = "mps"
+        override val licenseTier = LicenseTier.Free
+        // HEAD evidence: MPS archives use case-sensitive MPS-* filenames (mac adds a -macos infix).
+        override val urlFilenameTokens = listOf("MPS-")
+    }
+
     // ---- Google-published IDE ----
 
     /**
@@ -163,8 +241,11 @@ sealed interface IdeProduct {
 
     /**
      * Catch-all for any JetBrains product not in the well-known list. The caller
-     * supplies the public product `code` (e.g. "RR", "AC", "DG") and a license tier
-     * so paid/free policies still work. Everything else is descriptive metadata.
+     * supplies the public product `code` (e.g. "DS" for DataSpell, or "AC") and a license
+     * tier so paid/free policies still work. Everything else is descriptive metadata.
+     *
+     * This is the escape hatch for the products listed as deliberately out of scope next to
+     * [knownProducts] — reaching one does not require editing the catalog.
      */
     data class Custom(
         override val id: String,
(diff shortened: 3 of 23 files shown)


jonnyzzz:
## Review 2026-08-03

- **Banned pattern — runtime `assumeTrue` skip**: `ProductCatalogDocsDriftTest.kt:53,55` (`repoFile`) skips via `Assume.assumeTrue` when `mcp.repo.root` is unset or the target file is missing. Root CLAUDE.md bans detection-and-skip in tests (Gradle-task-level only). The line-55 skip is also self-defeating: if `website/content/docs/devrig.md` is moved/renamed, the drift guard silently passes — the exact drift #430's acceptance criterion requires to FAIL. Gradle always injects `mcp.repo.root` (build.gradle.kts), so both assumes should be hard failures (`error(...)`). ~3-line fix.
- **Minor — unpublished-platform message over-claims under `--version`**: `IdeReleaseLookup.kt` counts `candidateReleases`/`releasesOfferingDownloadKey` after the `wantedVersion`/`buildPrefix` filters, so asking for an old version that predates e.g. `linuxARM64` yields "$os/$arch is not a supported host for this product" even though current releases publish it and a version retry would help. Consider emitting this message only when no version/build filter is active.
- **Nit**: the `--product` help drift check (`ProductCatalogDocsDriftTest.kt:42`) matches ids anywhere in `Main.kt` source, so an id in any comment satisfies it — acceptable as a tripwire.
- **Verified sound**: merge-base is current main HEAD (30ef5a31, post-#412), MERGEABLE/CLEAN; the five new agent Dockerfiles are byte-identical to the current `clion-agent` pattern; `fromString` lowercases so `DB`/`DG`/`MPS` aliases all resolve; the DG→DB `installedProductCode` split matches the existing IIU→IU/PCP→PY shape and closes the prompts/AGENTS.md `DB` mismatch; download-list ordering test matches the tier grouping (Free → FFNC → Paid, catalog order); test helpers (`resolveArchiveFromFixtures`, `assertExpectedExtension`, `liveNetworkTest` task) all pre-exist. Offline matrix + opt-in live matrix + mutation-checked guards is honest coverage; new `unpublishedPlatformFailureMessage` correctly distinguishes missing-key from wrong-filename (filename rejects still increment `releasesOfferingDownloadKey` and fall to the old message).

**Verdict: needs-changes** — content and coverage are merge-ready and closes #430, but the `assumeTrue` skips in the new drift test are an explicitly banned pattern with a real silent-skip failure mode; trivial to fix, then good to merge.

jonnyzzz:
## Processed 2026-08-03

**Changes made on the branch before merging** (rebased onto `main` at bd4ae141, then two commits):

1. **Extended #431's all-products download guard** (`AllIdeProductsDownloadTest`), which landed on main after this branch was cut and asserts set equality against `IdeProduct.knownProducts`. Added the five new rows (rustrover, phpstorm, rubymine, datagrip, mps) mirroring the recorded `products-*.json` fixtures — including DataGrip's DG-query/DB-install code split and MPS's 261-baseline build — plus the matching `CLAUDE.md` prose fix ("all nine" → "every catalog product").
2. **Review blocker: `ProductCatalogDocsDriftTest` no longer skips.** The `Assume.assumeTrue` guards became hard failures: a missing `mcp.repo.root` (Gradle always sets it) throws, and a missing `website/content/docs/devrig.md` FAILS the drift guard instead of silently skipping it.
3. **Review nit: unpublished-platform diagnostic no longer over-claims.** When a `--version`/build filter excluded releases, the resolver error now names the active filter and says releases outside it may still cover the host, instead of declaring the platform "not a supported host for this product". Covered by a new test in `IdeReleaseLookupTest`.

**Tests run locally (sequentially, from an isolated worktree):**
- `./gradlew :intellij-downloader:test` — BUILD SUCCESSFUL (incl. `ProductCatalogDocsDriftTest` 2/2 and `IdeReleaseLookupTest` 36/36, zero skips)
- `./gradlew :npx-kt:test` — BUILD SUCCESSFUL (incl. `AllIdeProductsDownloadTest` 4/4 with the 14-product matrix)

**Not done here:** the five new Docker agent images (`phpstorm-agent`, `rubymine-agent`, `rustrover-agent`, `datagrip-agent`, `mps-agent`) were not built — that is TeamCity territory; the Dockerfiles follow the existing per-IDE pattern verbatim.

Merged into `main` via rebase (tip `50c973a1`); branch deleted.

PR #436, opened 2026-08-03
fix: deterministic output.jsonl order + fail-loud on write error (#284)

matveidetonator:
Two defects in ExecutionManager.responseBuilder: 
(1) each execution event was appended from its own launch on Dispatchers.IO, so output.jsonl lines raced into arbitrary order; 
(2) the fan-out ran under a SupervisorJob and build() only join()ed, so a real write failure was logged but the call still returned success (false ACK — #433 didn't cover this). Fix: new SerialWriteQueue (:execution-storage) drains writes in submission order and rethrows the first failure via awaitCompletion(); ExecutionManager routes all event paths through it and awaits in build(). Covered by SerialWriteQueueTest, an ExecutionStorageTest ordering case, and end-to-end ExecutionStorageTaskTest.


ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/execution/ExecutionManager.kt
```diff
@@ -218,7 +220,7 @@ class ExecutionManager(
             mcpProgress.report(text)
             _errorMessages.add(throwable.message ?: message)
 
-            innerScope.launch {
+            storageQueue.submit {
                 project.executionStorage.appendExecutionEvent(executionId, text)
```
jonnyzzz: project reference can leak into the queueu

jonnyzzz: so the better and more safe approach -- make the storageQueue be a per-project component, and so the submit { function gives you new project to each lambda. 

on the other hand, instead, we may keep only data, not the lambda in the queue

jonnyzzz: in the prev implementation the service was per project, so it disposes before the project is closed and thus project was not leaked (but captured)

execution-storage/src/main/kotlin/com/jonnyzzz/mcpSteroid/storage/SerialWriteQueue.kt
```diff
@@ -0,0 +1,63 @@
...
+ * [awaitCompletion]. When [parentScope] is cancelled, the worker is cancelled with it.
+ */
+class SerialWriteQueue(parentScope: CoroutineScope) {
+    // A supervisor child of parentScope: a failing write stays contained here (does not
+    // cancel parentScope) until awaitCompletion() re-raises it, yet parentScope cancellation
+    // still tears the worker down. Completed by awaitCompletion() so it never keeps
+    // parentScope structurally alive.
+    private val supervisor = SupervisorJob(parentScope.coroutineContext[Job])
+    private val scope = CoroutineScope(parentScope.coroutineContext + supervisor)
+    private val channel = Channel<suspend () -> Unit>(Channel.UNLIMITED)
```
jonnyzzz: unlimited channel can OOM in the IDE. I would limit it to 100 records. Once limited, the caller will have to suspend and wait for writer to complete. 

the old implementation has the same problem too

execution-storage/src/main/kotlin/com/jonnyzzz/mcpSteroid/storage/SerialWriteQueue.kt
```diff
@@ -0,0 +1,63 @@
...
+    private val worker = scope.async {
+        for (action in channel) {
+            action()
```
jonnyzzz: no exception handling here, so if something fails, an action fails, you stop processging the rest of the messages. It's more hard for you since messages are lambdas and you do not know what's inside. 

it's a bit inefficient, because you write each message indepdendently, each time dealing with files

execution-storage/src/main/kotlin/com/jonnyzzz/mcpSteroid/storage/SerialWriteQueue.kt
```diff
@@ -0,0 +1,63 @@
...
+        }
+    }
+
+    /**
+     * Enqueue an action to run after all previously submitted actions, on the queue's
+     * single worker. Non-suspending and ordered: the submission order is the execution
+     * order. Must not be called after [awaitCompletion].
+     */
+    fun submit(action: suspend () -> Unit) {
+        val result = channel.trySend(action)
```
jonnyzzz: this is not correct, there is `suspend`-ing version for channel which suspends on overflow.

now you are throwing exceptions to the clients when queue is full -- this breaking clients of your code, nobody expects logging to throw

ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/execution/ExecutionManager.kt
```diff
@@ -141,16 +142,17 @@ class ExecutionManager(
...
-            parentScope.coroutineContext +
-            storageJob +
-            Dispatchers.IO +
-            CoroutineName("storage-$executionId") +
-            ModalityState.any().asContextElement()
+        // Storage writes go through a single-worker queue so output.jsonl lines land in the
+        // exact order they were emitted (fan-out onto Dispatchers.IO used to scramble them,
+        // #284) and a genuine write failure surfaces from build() instead of being logged and
+        // ACKed as success (#433 follow-up). The queue owns a supervised child of this scope.
+        private val storageQueue = SerialWriteQueue(
```
jonnyzzz: I think all you want was to add ` withLimitedParallelism(1)` to the previous code, and it solves the problem. 

Executors and Dispatches preserve the FIFO on launch calls, the problme we had here was with N threads being used to dispatch events. 

Your tests (once changed to use this class) should be able to handle/check that

jonnyzzz (changes requested):

jonnyzzz (changes requested):
please review comments and it's good to go

ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/execution/ExecutionManager.kt
matveidetonator: had an agent write a probe test, order still breaks under limitedParallelism(1), and it doesn't fix the false ACK

ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/execution/ExecutionManager.kt
jonnyzzz: Reworked in 5c352bac: the queue now buffers plain data records (`ExecutionEventRecord.Append` / `CodeError` — executionId + text), never lambdas, so no `project` capture rides in the buffer. The builder resolves `project.executionStorage` once at queue construction; the worker stays a child of the per-call scope (cancelled with the call), so nothing outlives the tool call either. Kept it per-call rather than per-project: each call writes its own `output.jsonl` (ordering domain is per-execution), and per-call fail-loud at `build()` would otherwise need per-item completion tracking on a shared queue plus cross-call head-of-line blocking.

execution-storage/src/main/kotlin/com/jonnyzzz/mcpSteroid/storage/SerialWriteQueue.kt
jonnyzzz: Bounded in 5c352bac — `Channel(10_000)` instead of UNLIMITED, so memory is capped. One deliberate deviation from "limit to 100 and let the caller wait": overflow records a failure that fails the call loudly at `build()` instead of parking the producer. The emit API (`logOutput`/`logProgress`) is non-suspend and runs on script callback threads including the EDT — blocking there on disk backpressure is the modal-hang class of bug. And 100 would engage on healthy scripts: a tight 500-`println` loop outruns per-line appends immediately (the PR's own regression test is that workload). With batched draining the depth stays near zero, so the cap only bites when the disk is genuinely stuck — failing loud there is the honest outcome. `capacity` is a single constructor constant if you prefer blocking semantics after all.

execution-storage/src/main/kotlin/com/jonnyzzz/mcpSteroid/storage/SerialWriteQueue.kt
jonnyzzz: Both fixed in 5c352bac. Exception handling: each record's write is contained per-record (CancellationException rethrown first, per the hot-path rule); the FIRST failure is kept and re-raised by `awaitCompletion()` at `build()`, later ones log to stderr, and draining continues — previously the first throw killed the worker's for-loop and every later submit landed in a channel nobody read. Data records (not lambdas) are what make this safe to reason about. Efficiency: the worker now coalesces the buffered burst — a run of same-execution appends goes through `ExecutionStorage.appendExecutionEvents` as ONE file open (test pins exactly 2 writes for a 1+99 burst).

execution-storage/src/main/kotlin/com/jonnyzzz/mcpSteroid/storage/SerialWriteQueue.kt
jonnyzzz: `submit` no longer throws on any path (5c352bac): closed queue (late submit after `awaitCompletion()` or scope cancellation) logs to stderr and drops — too late to fail a call that already returned; full buffer records the overflow as the call's failure, surfaced at `build()`, never at the logging caller. The suspending `send` isn't reachable from here: the emit API is non-suspend (script callbacks, possibly EDT), so suspending would mean `trySendBlocking` — i.e. parking the emitting thread; see the capacity thread for why that's the worse trade.

ij-plugin/src/main/kotlin/com/jonnyzzz/mcpSteroid/execution/ExecutionManager.kt
jonnyzzz: Checked this carefully — the one-liner doesn't hold against this storage's internals. `appendExecutionEventJson` re-dispatches inside via `withContext(Dispatchers.IO)` (ExecutionStorage.kt), and a `limitedParallelism(1)` view's unit of limitation is a dispatch SEGMENT, not a coroutine: at that inner `withContext` the outer coroutine suspends, the view's single slot frees, the next launched coroutine starts, and both writes run concurrently on the unlimited IO pool — the scramble survives, just re-timed (the 500-line regression test catches it). FIFO-on-launch is true for dispatch order, but full-action ordering needs the write to run to COMPLETION before the next starts, which is what the single-consumer loop gives (it awaits each write across any internal re-dispatch). The launch-based shape also leaves no failure path to `build()` — `join()` never rethrows a child failure, supervisor or not, which was exactly the false-ACK half of this PR. So the repaired queue stays, but rebuilt per the other four comments: data records, bounded, drain-past-failure, batched, non-throwing submit — net simpler than the first version (no SupervisorJob/async dance).

execution-storage/src/main/kotlin/com/jonnyzzz/mcpSteroid/storage/SerialWriteQueue.kt
matveidetonator: in ca3fccb2 added batch cap, to prevent a burst that filled the buffer got joined into one string, peaking at
twice the pending bytes

matveidetonator:
ca3fccb2 fixes another bug:  cancelling a run threw away the end of its log. now it gets written on the way out and the test says so.

jonnyzzz:

